PR #20, opened 2021-04-23
Adding Pset_PavementCommon

larswik:


larswik:
We discovered the need for such a Pset when working with the machine control storyline.

IFC4x3/Properties/l/Length_0Cx6_hSdn2hR4RarQdjYBy/DocProperty.xml
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0" encoding="utf-8"?>
+<DocProperty xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" id="Length_0Cx6_hSdn2hR4RarQdjYBy" Name="Length" UniqueId="0cec6fab-727c-42ad-b11b-9356a7b622fc" PrimaryDataType="IfcPositiveLengthMeasure" />
```
SergejMuhic, line 2: I think the name Length is taken.

IFC4x3/Properties/s/StructuralSlope_0Neg5l7QLCGR0Sz5Gs_qdu/Documentation.md
```diff
@@ -0,0 +1 @@
+The nominal side slope (allowable steepness) of the pavement structure (not including side slope fill) as a ratio measure. The slope  information is provided in addition to the shape representation and the geometric parameters used within. In case of inconsistency between the geometric parameters take precedence.
```
SergejMuhic, line 1: Would be good to feature the boundary values of the ratio measure.

IFC4x3/Properties/t/ThicknessEnd_23IeYdlhfElwvDUY45W8gF/Documentation.md
```diff
@@ -0,0 +1 @@
+The nominal thickness of the pavement after a transition from its original value given as Thickness. The size information is provided in addition to the shape representation and the geometric parameters used within. In case of inconsistency between the geometric parameters and the size properties, provided in the attached property set, the geometric parameters take precedence. ThicknessEnd location shall be regarded as the pavement object’s placement plus its Length.
```
SergejMuhic, line 1: 'plus its Length' which Length is meant here?`Nominal or from geometric representation? Also, can it be clearly extracted from Representation? Just being careful for the sake of unambiguity.

SergejMuhic (commented):
Looks good.

IFC4x3/Properties/l/Length_0Cx6_hSdn2hR4RarQdjYBy/DocProperty.xml
larswik: I suggested this to Juha who commented in the following way: 
"I’m rather sceptical about the idea of defining attributes/properties taken out of their context, tends to become data rather than information as you clean the semantic meaning out of it.

If the problem for harmonizers is that the same property name appears in different PSets with different definitions, the appropriate solution in my mind would be more descriptive  naming rather that generalizing the definitions.

This length could be called NominalLength (the same in Psets having the same definition: course, kerbstone, precast kerb, transition section)? All these (and perhaps the couple of Psets that already have NominalLength) would need a small update in their definitions (none of the original ones seemed perfect, that’s why I proposed my own). 

The same could be done with thickness: NominalThickness with generic definition, the transition specific explanation could then be in NominalThicknessEnd".

Shall I change names? Have we used "Nominal..." as a prefix now in other places for Pset property names?

SergejMuhic:
Good thing for Github. I can see reverts already doing their job. :D

larswik:
Easier to close this pull request and create a new one for this property set than to try to figure out exactly which commits to revert...

PR #34, opened 2021-05-10
New proposal for a Pset_PavementCommon

larswik:


IFC4x3/Properties/n/NominalThicknessEnd_2dzrJAQ351fhj1A45yJpqq/Documentation.md
```diff
@@ -0,0 +1 @@
+The nominal thickness of the pavement after a transition from its original value given as Thickness. The size information is provided in addition to the shape representation and the geometric parameters used within. In case of inconsistency between the geometric parameters and the size properties, provided in the attached property set, the geometric parameters take precedence. NominalThicknessEnd location shall be regarded as the pavement object’s placement plus its nominal ength.
```
SergejMuhic, line 1: Is 'Thickness' supposed to be 'NominalThickness'?

SergejMuhic (commented):
Just one remark.

IFC4x3/Properties/n/NominalThicknessEnd_2dzrJAQ351fhj1A45yJpqq/Documentation.md
AlexBrad1eyCT: also "ength" at the end should be "length" !

IFC4x3/Properties/n/NominalThickness_3cTq6sD71BVAf9l72L6lIy/Documentation.md
```diff
@@ -0,0 +1 @@
+The nominal thickness of the pavement. The size information is provided in addition to the shape representation and the geometric parameters used within. In case of inconsistency between the geometric parameters and the size properties, provided in the attached property set, the geometric parameters take precedence. It is regarded as constant thickness of the pavement, unless property ThicknessEnd is provided, in which case Thickness shall be regarded as the value at the beginning of a transition (usually at the object placement location). Example of such case is a (road) transition segment where the pavement object’s linear placement along an alignment denotes the beginning location.
```
AlexBrad1eyCT, line 1: @larswik, is this a new property? is it worth generalising the NominalThickness property (so it can be used in other cases) and adding the Clarifications in the PSet or the Entity Documentation?

IFC4x3/Sections/Domain specific data schemas/Schemas/IfcRoad/PropertySets/Pset_PavementCommon/Documentation.md
```diff
@@ -0,0 +1 @@
+Properties for pavement.
```
AlexBrad1eyCT, line 1: As above possible location for clarifications on nominalThickness and possibly nominalLength.

AlexBrad1eyCT (changes requested):
Suggestions around generalisation of Nominal properties with clarifications within the Pset Documentation (this is probably the most suitable location) or the Entity Documentation (on the other hand this is the most visible location)

IFC4x3/Properties/n/NominalThicknessEnd_2dzrJAQ351fhj1A45yJpqq/Documentation.md
larswik: Fixed

IFC4x3/Properties/n/NominalThickness_3cTq6sD71BVAf9l72L6lIy/Documentation.md
larswik: I discussed with Juha H who didn't find the existing definitions suitable. However, if it's ok to generalize this and put any clarifications in the Pset documentation, I guess that could be ok (need to discuss with Juha first).

IFC4x3/Properties/n/NominalThickness_3cTq6sD71BVAf9l72L6lIy/Documentation.md
```diff
@@ -0,0 +1 @@
+The nominal thickness of the pavement. The size information is provided in addition to the shape representation and the geometric parameters used within. In case of inconsistency between the geometric parameters and the size properties, provided in the attached property set, the geometric parameters take precedence. It is regarded as constant thickness of the pavement, unless property NominalThicknessEnd is provided, in which case NominalThickness shall be regarded as the value at the beginning of a transition (usually at the object placement location). Example of such case is a (road) transition segment where the pavement object’s linear placement along an alignment denotes the beginning location.
```
SergejMuhic, line 1: Not sure how to fix the last sentence. @AlexBrad1eyCT any suggestion? I would probably just do e.g.

AlexBrad1eyCT: >  It is regarded as constant thickness of the pavement, unless property NominalThicknessEnd is provided......

slight rewording to:

> The thickness of the pavement remains constant, unless the property NominalThicknessEnd is provided. In which case NominalThickness is the value at the beginning of a transition (usually at the object placement location). e.g. A (road) transition segment where the pavement object's linear placement along an alignemnt denotes the begining location.

IFC4x3/Properties/n/NominalThickness_3cTq6sD71BVAf9l72L6lIy/Documentation.md
AlexBrad1eyCT: @larswik yes i understand Juha's point, but it conflicts with the contraints laid out from bSI... 2 options arise:

-  change the name to NominalPavementThickness, then the documentation can stand as is.
- set the property documentation to: 
> The nominal thickness of the object. The size information is provided in addition to the shape representation and the geometric parameters used within. In case of inconsistency between the geometric parameters and the size properties, provided in the attached property set, the geometric parameters take precedence.

Then add the rest of the clarification in the relation between properties to the PSet.

Actually in this particular example think the second option is better as we shouldnt really be referencing other properties within the property description. the impact of one property on another is at the scope level of the Pset therefore should be in the PSet documentation.

@SergejMuhic thoughts....

PS there could be a third option of a Complex Property but i think that just over complicates the issue...

AlexBrad1eyCT (approved):
LFTM, well done @larswik!

SergejMuhic (approved):

PR #45, opened 2021-05-26
Pset_GeotechnicalStratumCommon doc type-o fix

larswik:


larswik:
Fixes #46

SergejMuhic:
> 
> 
> Fixes #46

for automatic issue closure, Fixes has to be in the initial description. It can be then seen on the left pane under Linked issues. Otherwise, it apparently does not work.

PR #65, opened 2021-06-08
corrections of nbsp Error

AlexBrad1eyCT:
Fixing encoding error in XML when \&nbsp; should be \&amp;nbsp;


jmirtsch (approved):
Looks good to me, this highlights an issue for ifcdoc to improve the serialization so all documentation is written to markdown files and not embedded within xml.

AlexBrad1eyCT:
its the other way around @jmirtsch i added an issue to IfcDoc [ISSUE](https://github.com/buildingSMART/IfcDoc/issues/77) when it encodes in the XML it does not write the string in a html safe manner e.g. it writes `\&nbsp` instead of `\&amp;nbsp;` (the correct way)

SergejMuhic:
> 
> 
> yes @jmirtsch i added an issue to IfcDoc [ISSUE](https://github.com/buildingSMART/IfcDoc/issues/77) when it encodes in the XML it does not write the string in a html safe manner e.g. it writes `\&nbsp` instead of `\&amp;nbsp;` (the correct way)

Thanks @AlexBrad1eyCT . Thanks for the IfcDoc issue, we should definitely handle this on the IfcDoc side (at some point 😄 )

PR #68, opened 2021-06-11
mistake with markdown styling

### LINKED ISSUE #69: Markdown in IfcFixedReferenceSweptAreaSolid
Markdown not correct see PR #68 .

### LINKED ISSUE #70: documentation issues in IfcFixedReferenceSweptAreaSolid
some are addressed in PR #68 
@jmirtsch @SergejMuhic @PeterBonsma (not watching yet)

a few more: 

> EXAMPLE  The reference surface is any surface (plane, cylindric, composite) situated in 3D space and positioned in the object coordinate system. In many cases, it is a surface of extrusion. The directrix lies on the surface, often defined as a p-curve at this reference surface. At any point of the directrix, a plane can be constructed. The origin of the position coordinate system lies at the directrix. The Axis3 (the z-axis, or normal) of the position coordinate system is identical to the tangent of the directrix at this point, the Axis1 (the x axis, or u) of the position coordinate system is identical to the FixedReference direction. The Axis2 (the y axis, or v) is constructed. In this case the resulting swept solid is not repositioned.

seems to be wrong here -> looks as it belongs to IfcSurfaceCurveSweptAreaSolid

> NOTE  In the case of a curve composed of IfcCurveSegment, the IfcSweptAreaSolid Position is dependent on the IfcCurveSegment StartPlacement and the type of BasisCurve which defines the name-function between the start and end (explicitly defined as the start point of the next segment) points of the segment

should be double checked with the last definition of IfcCurveSegment (e.g. it is called ParentCurve there, not BaseCurve).

also the informal proposition number 3 says: The Directrix curve shall be tangent continuous. This would also apply to IfcCurveSegment. Here all IfcCurveSegments (beside the last) shall have the Transition CONTSAMEGRADIENT or CONTSAMEGRADIENTSAMECURVATURE. Maybe this could be added as a formal or informal proposision

finally we should add a figure, similar to the one in ISO 13030-42

TLiebich:
wrong escape of * - like \*
please check ifcDoc export since the \* characters are all over

Fixes #69 

added Improved definition of IfcFixedReferenceSweptAreaSolid to clarify the use of FixedReference
(should have been a separate pull request, but seems to be added here. (please double check)

Fixes #70


pjanck (approved):
lftm

AlexBrad1eyCT (commented):
LFTM, are we going to add the corrections to be discussed in #70, ? Or shall we detach the issue and put those changes in a new pull request? Will approve on decision

jmirtsch (approved):
Looks good.  I have searched the remainder of the spec for /*/* and included fixes in #81

SergejMuhic (approved):
Merging.

PR #77, opened 2021-06-17
IfcDoc Serialization Consistent Partial Templates

jmirtsch:
These changes should permit consistent concept template serialization with associated IfcDoc changes.


IFC4x3/Templates/Partial Templates/Quantities/DocTemplateDefinition.xml
```diff
@@ -7,13 +7,5 @@
 			</Rules>
 		</DocModelRuleAttribute>
 	</Rules>
```
SergejMuhic, line 9: Was thinking about this. Does it make sense to completely remove these "Templates" nodes when writing the files? I guess the way it is for now I merge. But these hrefs look okay actually.

SergejMuhic (commented):
lftm. merging.

PR #84, opened 2021-06-22
Restore Examples revised to IFC4.3RC3

jmirtsch:
Examples were disabled during Release Candidate publishing.  They were not compliant with recent versions of IFC.
I've revised them according to IFC4.3RC3 and validated them with the KIT checker.  

Alignment examples (and other infrastructure examples) will need to be revised in a separate task.


pjanck:
Before I start reviewing these files, I'd propose to wait for RC4 and I'll try and run them through TUM OpenInfraPlatform with a script. I can then check against the screen shots from 4x1 version of the standard and prove:

- files have no syntax mistakes (my parser will complain)
- files can be visualized (at least that there are no mistakes in geometry)

I cant vouch for other content, though. What do you think?

SergejMuhic:
Maybe could run the IfcKit over this also. I will postpone this a bit then.

PR #100, opened 2021-06-27
Found a couple of missing indentations

SergejMuhic:
Indentations for 'NOTE' missing. Added '&gt;'


IFC4x3/Sections/Resource definition data schemas/Schemas/IfcQuantityResource/Entities/IfcQuantityWeight/DocEntity.xml
```diff
@@ -17,7 +17,7 @@
 	</Attributes>
 	<WhereRules>
 		<DocWhereRule Name="WR21" UniqueId="eedb5c99-fdbd-458f-adb9-eac39d641365">
-			<Documentation>If a unit is given, the unit type shall be mass unit. NOTE&amp;nbsp; There is no distinction between the concept of &quot;Mass&quot; and &quot;Weight&quot; in the current IFC Release.</Documentation>
+			<Documentation>If a unit is given, the unit type shall be mass unit. &gt;NOTE&amp;nbsp; There is no distinction between the concept of &quot;Mass&quot; and &quot;Weight&quot; in the current IFC Release.</Documentation>
```
pjanck, line 20: Other documentation tags have a line break before the NOTE - is it intentional / handled differently here?

IFC4x3/Constants/h/HELMERTCURVE_14V3RUc6LFh9bZo1rEFhxL.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <DocConstant xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" id="HELMERTCURVE_14V3RUc6LFh9bZo1rEFhxL" Name="HELMERTCURVE" UniqueId="447c36de-9865-4fac-9963-c81d4e3ebed5">
 	<Documentation>Helmert transition.\
```
pjanck, line 3: What's the point of the backslash at the end? It seems out of place.

pjanck (commented):
Some questions (perhaps trivial).

IFC4x3/Constants/h/HELMERTCURVE_14V3RUc6LFh9bZo1rEFhxL.xml
pjanck: It isn't used in any of the other changed descriptions in this PR. I'd prefer to see consistency, but can leave it as is.

SergejMuhic: Thing is, we never agreed on guidelines. I have proposed to @jmirtsch to use something visible instead of two spaces and we found the backslash option. The change is coming from @czapplitec though, not me. 😄

pjanck (approved):
lftm

PR #120, opened 2021-07-02
PW Schema Documentation and rename to convention

AlexBrad1eyCT:
Added Schema Documentation and rename schema to convention

Renaming has caused movement of folder contents only true changes are to Documentation.md and the where rules on PW entities


larswik (approved):

SergejMuhic (approved):
This would have to be done for road and rail also. I will add @jmirtsch analysis to an issue for the liaisons to use.

See #122

PR #125, opened 2021-07-03
Wrong schema reference in description

SergejMuhic:
IfcSharedBuildingElements instead of IfcSharedBldgElements


AlexBrad1eyCT (approved):
LFTM

PR #136, opened 2021-07-04
Save after documentation generation and PR closing

SergejMuhic:
Mostly it is about the added changes set

Need to re-base the local forks on this as there is a new set of change sets


AlexBrad1eyCT (approved):
LFTM

Issue #138, opened 2021-07-06
_Axis_ again. Should this instead be "Supported shape representations for IfcAlignment are (RepresentationIdentifier='Axis" and RepresentationType="Curve2D/Curve3D"):


SergejMuhic:
_Axis_ again. Should this instead be "Supported shape representations for IfcAlignment are (RepresentationIdentifier='Axis" and RepresentationType="Curve2D/Curve3D"):

Address after merging of #130
...

_Originally posted by @larswik in https://github.com/bSI-InfraRoom/IFC-Specification/pull/130#discussion_r663775125_

PR #146, opened 2021-07-10
IfcAlignment documentation

larswik:
Adresses #138

Fixed some small type-o's and replaced IfcAlignment.Axis with shape representation.
I am unsure if it is clear enough to reference "shape representation" or if a reference to the "Alignment geometry" template should be added?


SergejMuhic (approved):
Looks good.

jmirtsch (approved):

Issue #108, opened 2021-06-30
WP3-PropertyHarmonization Road Properties


christophe-blt:
You will find the HarmonizationExcel here: 
[HarmonizationCases_30-06-2021.xlsx](https://github.com/bSI-InfraRoom/IFC-Specification/files/6739776/HarmonizationCases_30-06-2021.xlsx)

1. Pset_PavementCommon -> Property Reference_00pQO0qSSHuO00025QrEV should be merged with other reference property (for instance Reference_1LtTm0qWWHuO00025QrEV  **HarmonizationCase1.2 =30**

2. Pset_PavementCommon -> NominalLength_0NS18P1J51kRFniH3h5CIL . Is it possible to merge it with NominalLength_07RHK0qUaHuO00025QrEV ? (It means changing the Documentation and DataType to IfcNonNegativeLengthMeasure instead of IfcPositiveLengthMeasure) Or change the property Name.  **HarmonizationCase1=23**

3. Pset_KerbStone;Pset_OnSiteCastKerb;Pset_PrecastKerbStone -> NominalHeight_1Fy4XYaJXDTBDDb9f06ofi, need to change the name or change the Documentation and Datatype to harmonize it with NominalHeight_1_HEQ0qSOHuO00025QrEV. **HarmonizationCase1=42**

4. Pset_SlabCommon;Pset_PavementCommon -> EnumeratedValue  Status_002JG0qWWHuO00025QrEV should be merged with Status_2OUyO0qS0HuO00025QrEV.   **HarmonizationCase1.2 =189**

5. Pset_PavementCommon -> NominalThickness_3cTq6sD71BVAf9l72L6lIy harmonize documentation with NominalThickness_2QgJLpWOb3xfvwr_YOT_k. **HarmonizationCase1.1 =49**

6. Pset_PlantCommon -> Description_3Zrq848Q59lhSMKyRwz2po. Do we need a Description property, it's already an IFC Class attribute. If it's needed, can change the name of the property and add documentation. 

7. Pset_RoadDesignCriteriaCommon -> There is no Applicable Type to this property 
![image](https://user-images.githubusercontent.com/82803096/123938502-7750b880-d997-11eb-9eab-c3cf3a8c4e0b.png)

Please let me know if i miss a pull request and this list is not relevant anymore.

PR #236, opened 2021-08-26
Deal with #108 (parts - see discussion)

larswik:


AlexBrad1eyCT (approved):
lftm

SergejMuhic (approved):
Just a note to remove the orphans when publishing.

CC @jmirtsch

PR #268, opened 2021-10-05
Correcting Body Constraint & errors in other geometry Identifiers

### LINKED ISSUE #252: Child templates of Body geometry
Child templates do not have the RuleID `Identifier` specified as in the top node `Body Geometry`
```
<Constraint Expression="Identifier[Value] = 'Body'" />
```
According to my knowledge, the inheritance is not a feature between concept templates.

CC @MatthiasWeise asking for some input to create a PR.

AlexBrad1eyCT:
Fixes #252

adds missing Body representationIdentifier value constraints on child elements.
Discovered errors, inconsistancies and missing identifiers in other geometry CTs so addresses these.

I ran a local documentation generation an the updates fix the issues in the mvdXML and the diagrams

### DIFF
--- a/IFC4x3/Templates/Product Shape/Product Geometric Representation/Body Geometry/Body AdvancedBrep Geometry/DocTemplateDefinition.xml
+++ b/IFC4x3/Templates/Product Shape/Product Geometric Representation/Body Geometry/Body AdvancedBrep Geometry/DocTemplateDefinition.xml
@@ -16,7 +16,7 @@
 										</DocModelRuleAttribute>
 										<DocModelRuleAttribute CardinalityMin="1" CardinalityMax="1" Name="RepresentationIdentifier" Identification="Identifier">
 											<Rules>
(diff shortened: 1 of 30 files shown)


jmirtsch:
Hi Alex, how did you generate this?  Seems Unique Ids weren't assigned as "IfcLabel_0000000000000000000000" is serialized multiple times.

AlexBrad1eyCT:
@jmirtsch, I used IfcDoc ,so it must be in the object creation or serialisation. I thought it was a bit weird, but the generated mvdXML in the documentation output seemsuneffected and correct

SergejMuhic (commented):
Looks good but we seem to have a problem with IfcDoc serialization in that it adds an id. Not sure why this was added, it has these tags in IfcDoc:
[IgnoreDataMember] [XmlIgnore]

@jmirtsch before merging, we have to talk about this.

jmirtsch (changes requested):
Fix UniqueIds

PR #283, opened 2021-11-03
Ifc doc read write

jmirtsch:
IfcDoc serializes xml id without a $ sign, and there are some other "formatting" changes that don't conform to IFC serialization.
This pull request should not impact schema.  Orphaned constant "PLAINTRACKSUPERSTRUCTURE" is removed.


AlexBrad1eyCT (approved):
All Good Jon, thanks for the corrections

PR #284, opened 2021-11-03
Remove Partial Template ModelRuleEntity duplication

jmirtsch:
I've taken a look at some duplication of DocModelRuleEntity within partial templates.
I can't ascertain exactly when and why this happened, I believe they are inherited from original baselines.
I've manually removed the duplication I identified in this pull request.


AlexBrad1eyCT (approved):
LFTM, just corrections to serialisation errors. @SergejMuhic if you want to check it merge straight after please

PR #291, opened 2021-11-08
Syncing templates of MVDs

AlexBrad1eyCT:
syncing the templates in reference view and other views that are not in the spec currently


pjanck (approved):
lftm

PR #330, opened 2021-12-12
Feature/georeferencing

### LINKED ISSUE #83: [IFC-Tunnel] Georeferencing re-work
The document summarizing the proposal:
**[DOCUMENTATION]** [geo_ref.pdf](https://github.com/bSI-InfraRoom/IFC-Specification/files/6791367/geo_ref.pdf)

Due to some requirements a re-work of georeferencing is required.

1. Non projected coordinate transformations
2. Support for geographic coordinate systems
3. Local placement reference to a geometric representation context (not in scope of IFC Tunnel for now)

SergejMuhic:
Resolves #83 

Added:
1. `IfcGeographicCRS`
2. `IfcRigidOperation`
3. `IfcWellKnownTextCRS`

ping @LeeGregory12d


IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcGeographicCRS/Documentation.md
```diff
@@ -0,0 +1 @@
+_IfcGeographicCRS_ is a coordinate reference system that is based on a geoid or ellipsoid. Geographic CRS typically requires two angles and a height (can be derived) to specify a location on a body such as ellipsoid, geoid etc
```
pjanck: Please delete geoid. This establishes (normal) height and uses an ellipsoid as base for localization. It is not part of / type of a *geographic* CRS.

Geographic CRS's height is "height above/below the ellipsoid's surface".

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcTranslationOperation/DocEntity.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<DocEntity xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" Name="IfcTranslationOperation" UniqueId="946a945b-8a4a-4e2c-9840-b56032727b5f" BaseDefinition="IfcCoordinateOperation" EntityFlags="32">
+	<Attributes>
+		<DocAttribute Name="Eastings" UniqueId="f2534a2f-e119-4a52-b1cf-a55c170eea45" DefinedType="IfcLengthMeasure" />
+		<DocAttribute Name="Northings" UniqueId="366b1fda-0ecf-4891-abb5-212d322b50fb" DefinedType="IfcLengthMeasure" />
+		<DocAttribute Name="OrthgonalHeight" UniqueId="caad8793-db34-4377-b751-b1591d8f3fe3" DefinedType="IfcLengthMeasure" AttributeFlags="1" />
```
pjanck: The naming is debatable, the types are weird.

- Easting/Northings *implies* UTM projection in some cases. Definitely not the best solution.
- Additionally, I thought that this entity could be used with `IfcGeographicCRS` - `IfcLengthMeasure` is false then, since the CRS uses angles.
- Typo `Orthogonal` 
- although I'd remove the prefix `Orthogonal` altogether and only use *Height* - it depends on the underlying `IfcCoordinateReferenceSystem` what this height is.
- What does `AttributeFlags` mean at height?

Proposal: 
- use `Axis1`, `Axis2` and `Axis3` as names for versality reasons. Or something similar.
- for type: use `IfcMeasureValue` with a WHERE rule checking for correct typing.

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcTranslationOperation/Documentation.md
```diff
@@ -0,0 +1 @@
+A translation operation specifies an offset in the coordinate reference system. It does not specify any conversion or distortions only a translation.
```
pjanck: I'd add an additional comment:

With it, only the point of origin of the context is defined; the coordinate axes of the context are not necessarily equal to the underlying CRS and are defined with ...

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcWellKnownTextCRS/Documentation.md
```diff
@@ -0,0 +1 @@
+_IfcWellKnownTextCRS_ is a coordinate reference system that is based on a standardized ISO format (ISO 19182).
```
pjanck: ISO 19162 (six, not eight!)

pjanck (changes requested):
Good starting point. Let's discuss.

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/DocSchema.xml
```diff
@@ -52,6 +52,8 @@
 				<DocDefinitionRef Name="IfcReal" UniqueId="53666384-7ea3-46d2-ba10-986fd16883d8" />
 				<DocDefinitionRef Name="IfcNamedUnit" UniqueId="04a55c7c-d817-4574-83c4-325bf83bbd76" />
 				<DocDefinitionRef Name="IfcLogical" UniqueId="155b841f-664f-4175-818f-326f654bf3fe" />
+				<DocDefinitionRef Name="IfcValue" UniqueId="ebb6f169-0bb0-4d4c-adf3-b93a7b706ca5" />
+				<DocDefinitionRef id="IfcMeasureValue_0zhETP7Hz2E94mtLjvV20s" Name="IfcMeasureValue" UniqueId="3dace759-1d1f-4238-9130-dd5b797c2036" />
```
pjanck, line 57: How come the rest have no `id` attribute?

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcGeographicCRS/Documentation.md
```diff
@@ -1 +1 @@
-_IfcGeographicCRS_ is a coordinate reference system that is based on a geoid or ellipsoid. Geographic CRS typically requires two angles and a height (can be derived) to specify a location on a body such as ellipsoid, geoid etc
+_IfcGeographicCRS_ is a coordinate reference system that is based on a ellipsoid. Geographic CRS typically requires two angles and a height to determine a unique position.
```
pjanck:
```suggestion
_IfcGeographicCRS_ is a coordinate reference system that is based on an ellipsoid. Geographic CRS typically requires two angles and a height above the ellipsoid's surface to determine a unique position in 3D space.
```

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcTranslationOperation/DocEntity.xml
```diff
@@ -1,9 +1,33 @@
...
+			</Definition>
+		</DocAttribute>
+		<DocAttribute Name="SecondCoordinate" UniqueId="366b1fda-0ecf-4891-abb5-212d322b50fb" DefinedType="IfcMeasureValue">
+			<Documentation>The second coordinate of translation. Can be a length measure in case of map coordinates and angle in case of geographic reference systems.</Documentation>
+			<Definition>
+				<DocDefinitionRef xsi:nil="true" href="IfcMeasureValue_0zhETP7Hz2E94mtLjvV20s" />
+			</Definition>
+		</DocAttribute>
+		<DocAttribute Name="Height" UniqueId="caad8793-db34-4377-b751-b1591d8f3fe3" DefinedType="IfcLengthMeasure" AttributeFlags="1">
+			<Documentation>Height above the coordinate plane.</Documentation>
```
pjanck:
```suggestion
			<Documentation>Height above (positive) or below (negative) the coordinate surface.</Documentation>
```

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcTranslationOperation/DocEntity.xml
```diff
@@ -1,9 +1,33 @@
 <?xml version="1.0" encoding="utf-8"?>
 <DocEntity xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" Name="IfcTranslationOperation" UniqueId="946a945b-8a4a-4e2c-9840-b56032727b5f" BaseDefinition="IfcCoordinateOperation" EntityFlags="32">
 	<Attributes>
-		<DocAttribute Name="Eastings" UniqueId="f2534a2f-e119-4a52-b1cf-a55c170eea45" DefinedType="IfcLengthMeasure" />
-		<DocAttribute Name="Northings" UniqueId="366b1fda-0ecf-4891-abb5-212d322b50fb" DefinedType="IfcLengthMeasure" />
-		<DocAttribute Name="OrthgonalHeight" UniqueId="caad8793-db34-4377-b751-b1591d8f3fe3" DefinedType="IfcLengthMeasure" AttributeFlags="1" />
+		<DocAttribute Name="FirstCoordinate" UniqueId="f2534a2f-e119-4a52-b1cf-a55c170eea45" DefinedType="IfcMeasureValue">
+			<Documentation>The first coordinate of translation. Can be a length measure in case of map coordinates and angle in case of geographic reference systems.</Documentation>
```
pjanck:
```suggestion
			<Documentation>The first coordinate of translation. Can be a length measure in case of map coordinates and solid angle measure in case of geographic coordinates.</Documentation>
```

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcTranslationOperation/DocEntity.xml
```diff
@@ -1,9 +1,33 @@
...
+			<Definition>
+				<DocDefinitionRef xsi:nil="true" href="IfcMeasureValue_0zhETP7Hz2E94mtLjvV20s" />
+			</Definition>
+		</DocAttribute>
+		<DocAttribute Name="SecondCoordinate" UniqueId="366b1fda-0ecf-4891-abb5-212d322b50fb" DefinedType="IfcMeasureValue">
+			<Documentation>The second coordinate of translation. Can be a length measure in case of map coordinates and angle in case of geographic reference systems.</Documentation>
```
pjanck:
```suggestion
			<Documentation>The second coordinate of translation. Can be a length measure in case of map coordinates and solid angle measure in case of geographic coordinate.</Documentation>
```

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcTranslationOperation/DocEntity.xml
```diff
@@ -1,9 +1,33 @@
...
+				<DocDefinitionRef xsi:nil="true" href="IfcMeasureValue_0zhETP7Hz2E94mtLjvV20s" />
+			</Definition>
+		</DocAttribute>
+		<DocAttribute Name="Height" UniqueId="caad8793-db34-4377-b751-b1591d8f3fe3" DefinedType="IfcLengthMeasure" AttributeFlags="1">
+			<Documentation>Height above the coordinate plane.</Documentation>
+		</DocAttribute>
 	</Attributes>
+	<WhereRules>
+		<DocWhereRule Name="CoordianteType" UniqueId="e9bb63a8-0676-42b5-93db-fa2803a66ae2">
+			<Documentation>Restricts the value type of _FirstCoordinate_ and _SecondCoordinate_ to _IfcLengthMeasure_ and _IfcPositivePlaneAngleMeasure_.</Documentation>
```
pjanck:
```suggestion
			<Documentation>Restricts the value type of _FirstCoordinate_ and _SecondCoordinate_ to _IfcLengthMeasure_ and _IfcPlaneAngleMeasure_.</Documentation>
```

Can be negative for western hemisphere.

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcTranslationOperation/DocEntity.xml
```diff
@@ -1,9 +1,33 @@
...
+			<Expression>(&apos;IFCMEASURERESOURCE.IFCLENGTHMEASURE&apos; IN TYPEOF(FirstCoordinate)
+OR
+&apos;IFCMEASURERESOURCE.IFCPOSITIVEPLANEANGLEMEASURE&apos; IN TYPEOF(FirstCoordinate))
```
pjanck:
```suggestion
&apos;IFCMEASURERESOURCE.IFCPLANEANGLEMEASURE&apos; IN TYPEOF(FirstCoordinate))
```

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcTranslationOperation/DocEntity.xml
```diff
@@ -1,9 +1,33 @@
...
+AND
+(&apos;IFCMEASURERESOURCE.IFCLENGTHMEASURE&apos; IN TYPEOF(SecondCoordinate)
+OR
+&apos;IFCMEASURERESOURCE.IFCPOSITIVEPLANEANGLEMEASURE&apos; IN TYPEOF(SecondCoordinate))</Expression>
```
pjanck:
```suggestion
&apos;IFCMEASURERESOURCE.IFCPLANEANGLEMEASURE&apos; IN TYPEOF(SecondCoordinate))</Expression>
```

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcTranslationOperation/Documentation.md
```diff
@@ -1 +1 @@
-A translation operation specifies an offset in the coordinate reference system. It does not specify any conversion or distortions only a translation.
+A translation operation specifies an offset in the coordinate reference system. It does not specify any conversion or distortions only a translation. The coodinate axes of the _IfcGeometricRepresentationContext_ can rotate and set a new basis coordinate system.
```
pjanck:
```suggestion
A translation operation specifies an offset in the coordinate reference system. It does not specify any transformation (conversion or distortions) only a translation. The coordinate axes of the _IfcGeometricRepresentationContext_ can rotate and set a new basis coordinate system.
```

pjanck: I need to think about this definition a bit more; I'm not yet happy with it.

pjanck (changes requested):

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcGeographicCRS/DocEntity.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<DocEntity xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" Name="IfcGeographicCRS" UniqueId="2d6be663-8d5e-42cc-b5b4-13bb37f41262" BaseDefinition="IfcCoordinateReferenceSystem" EntityFlags="32">
+	<Attributes>
+		<DocAttribute Name="GeodeticDatum" UniqueId="b4d581f7-612a-4e98-963a-95ad06714397" DefinedType="IfcIdentifier" AttributeFlags="1" />
+		<DocAttribute Name="Unit" UniqueId="82bf71dc-4f74-483b-84c7-70c84a5ce22d" DefinedType="IfcNamedUnit" AttributeFlags="1" />
```
pjanck: Proposal: I imagine this entity as:

grafik

pjanck (changes requested):
Wouldn't it make more sense to add a WKT attribute to both IfcProjectedCRS and IfcGeographicCRS to semantically distinguish the type of CRS without the need to look into the WKT string itself? Otherwise, the software *must* peek into the WKT string and understand it ...

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcTranslationOperation/DocEntity.xml
```diff
@@ -1,9 +1,33 @@
...
+			<Definition>
+				<DocDefinitionRef xsi:nil="true" href="IfcMeasureValue_0zhETP7Hz2E94mtLjvV20s" />
+			</Definition>
+		</DocAttribute>
+		<DocAttribute Name="Height" UniqueId="caad8793-db34-4377-b751-b1591d8f3fe3" DefinedType="IfcLengthMeasure" AttributeFlags="1">
+			<Documentation>Height above the coordinate plane.</Documentation>
+		</DocAttribute>
 	</Attributes>
+	<WhereRules>
+		<DocWhereRule Name="CoordianteType" UniqueId="e9bb63a8-0676-42b5-93db-fa2803a66ae2">
```
pjanck:
```suggestion
		<DocWhereRule Name="CoordinateType" UniqueId="e9bb63a8-0676-42b5-93db-fa2803a66ae2">
```

pjanck (changes requested):
Are the attributes *FirstCoordinate* and *SecondCoordinate* optional? If yes, the Where rule should account for that.

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcWellKnownTextCRS/Documentation.md
```diff
@@ -0,0 +1 @@
+_IfcWellKnownTextCRS_ is a coordinate reference system that is based on a standardized ISO format (ISO 19162).
```
anborr:
```suggestion
_IfcWellKnownTextCRS_ is a coordinate reference system that is defined by a text string whose format is specified in ISO 19162.
```

pjanck:
Additional suggestion: rename `IfcGeographicCRS` to `IfcGeodeticCRS` to be in line with ISO 19111:2020.

EDIT 2022-01-18: Don't.

anborr:
In the most common use cases, we want to either define (1) a local undistorted coordinate system, a.k.a. “engineering coordinate system” whose origin is located by referring to a mapped Coordinate Reference System (Projected CRS) , or (2) a distorted project coordinate system that uses the same mapping parameters as the Projected CRS, but has an offset (=translation). This is often necessary in large infrastructure projects. Lee Gregory is referring to this as “truncated coordinate system”: https://github.com/buildingSMART/IFC4.3.x-development/issues/43 

Case (1) seems to be covered by IfcMapConversion where SourceCRS is referring to the local engineering coordinate system, represented by the IfcGeometricRepresentationContext. IfcMapConversion has scale (in the newest version also in y and z direction). Using these scale factors essentially means that we describe the (local) distortion of the **target** (!) coordinate system. There are two issue with that: (A) @pjanck is right that the distortion is already implicitly defined by the ProjectedCRS – so we have redundancy with the risk of inconsistency. (B) As the local coordinate system is undistorted in case (1), the users might assume that scale must equal 1. However, as we must define the transformation **from** reality (or a model with real-world dimensions) **into** the ProjectedCRS, it must essentially be the **inverse** local distortion factor of the projected CRS. This should be made much clearer in the documentation. A scale factor = 1 would mean the projected CRS to be undistorted. This exists only at the touching lines of the Transverse Mercator projection. Most users will assume that the scale factor describes the scale of their model wrp. to reality and define the scale=1 to specify it as being an undistorted 1:1 model. Only a few will understand that we must decribe the the characteristics of the ProjectedCRS at this point, and even less will be able to specify it correctly.
 
With this pull request, case (2) can now be covered by IfcTranslationOperation in combination with IfcProjectedCRS as TargetCRS. The current documentation reads “It does not specify any conversion or distortions only a translation.” However, this is misleading as the translation applied to a distorted (mapped) coordinate system will result in a distorted local coordinate system (which exactly is the goal of case (2) or ‘truncating’ coordinates).

As the meaning of translation/projection/mapping is very counter-intuitive in the current approach, we should consider to take a more radical approach and describe the local Coordinate System as a transformation *from reality*, instead of the other way round as it is currently implemented in IFC: “IfcGeometricRepresentationContext has an optional inverse attribute HasCoordinateOperation pointing to an IfcMapConversion. This implies that a coordinate operation has to be performed to convert coordinates **to** the projected CRS (map coordinates).”
Taking the other direction and describing the used coordinate system as a transformation *from* reality would comply better with the approach taken by the Geodesy/Geoinformatics community and the corresponding international standards.

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcGeographicCRS/Documentation.md
```diff
@@ -0,0 +1 @@
...
```
anborr, line 1: A geographic coordinate reference system is a CRS that uses a three-dimensional spherical surface to determine locations on the Earth. Any location on Earth can be referenced by a point with longitude and latitude coordinates.

anborr:
```suggestion
A _IfcGeographicCRS_ is a CRS that uses a three-dimensional ellipsoid surface to determine locations on the Earth. Any location on Earth can be referenced by a point with longitude and latitude coordinates.
```

anborr:
Would using `IfcTranslationOperation` with TargetCRS = `IfcGeographicCRS` mean that the all coordinates in the IFC model are latitude/longitude?

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcGeographicCRS/Documentation.md
anborr: grafik

anborr:
> Additional suggestion: rename `IfcGeographicCRS` to `IfcGeodeticCRS` to be in line with ISO 19111:2020.

Might make sense as there is a slight difference between **geodetic** and **geocentric**:

From [Wikipedia](https://en.wikipedia.org/wiki/Geodetic_coordinates#Geodetic_vs._geocentric_coordinates): 
"Geodetic latitude and geocentric latitude have different definitions. Geodetic latitude is defined as the angle between the equatorial plane and the surface normal at a point on the ellipsoid, whereas geocentric latitude is defined as the angle between the equatorial plane and a radial line connecting the centre of the ellipsoid to a point on the surface (see figure). When used without qualification, the term latitude refers to geodetic latitude. For example, the latitude used in geographic coordinates is geodetic latitude. The standard notation for geodetic latitude is φ. There is no standard notation for geocentric latitude; examples include θ, ψ, φ′.

Similarly, geodetic altitude is defined as the height above the ellipsoid surface, normal to the ellipsoid; whereas geocentric altitude is defined as the distance to the reference ellipsoid along a radial line to the geocenter. When used without qualification, as in aviation, the term altitude refers to geodetic altitude (possibly with further refinements, such as in orthometric heights). Geocentric altitude is typically used in orbital mechanics (see orbital altitude). "

anborr:
The documentation of `IfcCoordinateOperation` (abstract super class IfcMapConversion / IfcTranslationOperation) reads:

"The coordinate operation is an abstract supertype to handle any operation (transformation or conversion) between two coordinate reference systems. It is meant to provide expandability for future versions, since currently only the conversion of a local engineering coordinate system into a map coordinate reference system is dealt with by the subtype _IfcMapConversion_."

This documentation must be changed.

Use cases, where the `SourceCRS` of `IfcCoordinateOperation` is not the local coordinate system of the IFC Model (`IfcGeometricRepresentationContext`), but another CRS seem not logical (unlikely to exist). Hence `IfcCoordinateReferenceSystemSelect` should be replaced by a direct reference to `IfcGeometricRepresentationContext`.

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcTranslationOperation/Documentation.md
anborr: “It does not specify any transformation (conversion or distortions) only a translation.” This is misleading as the translation applied to a distorted (mapped) coordinate system will result in a distorted local coordinate system (which exactly is the goal of ‘truncating’ coordinates). In addition, a translation *is* a transformation. Moreover, a rotation is not covered by translation.

anborr:
WHERE rule must be added to IfcMapConversion ta make sure that TargetCRS is **not** of type IfcGeographicCRS (as mapping means projection and a GeographicCRS is not a mapped CRS).

anborr:
IfcMapConversion: We should consider to entirely remove the scale attributes as they are misleading, see https://github.com/bSI-InfraRoom/IFC-Specification/pull/330#issuecomment-1009362066

If not, at least the documentation of IfcMapConversion must be revised. 

> Scale: Scale to be used, when the units of the CRS are not identical to the units of the engineering coordinate system. If omited, the value of 1.0 is assumed.

It's not a question of units! The scale factor is a characteristeric of the mapping.

> Scale Y by which the length measures in local engineering CS have to be multiplied to get map coordinates.

It's not so easy as we will have to combine it with the offset (=translation).

Same for `ScaleZ`. Indeed, ScaleZ might not exist in reality, see https://github.com/buildingSMART/IFC4.3.x-development/issues/43 

By contrast, we definitly need Scale Y, as we will very often have differences in ScaleX and ScaleY, since the distortation occurs only in one direction in Transversal Mercator projections.

anborr:
> Wouldn't it make more sense to add a WKT attribute to both IfcProjectedCRS and IfcGeographicCRS to semantically distinguish the type of CRS without the need to look into the WKT string itself? Otherwise, the software _must_ peek into the WKT string and understand it ...

I agree.

anborr (changes requested):

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/DocSchema.xml
SergejMuhic: Good question. Have to look into it.

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcGeographicCRS/DocEntity.xml
SergejMuhic: Is `PrimeMeridian` still a thing? https://github.com/pjanck/Georef-IFC4x3-rework/issues/28

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcTranslationOperation/DocEntity.xml
SergejMuhic: Maybe I reference the exact defined types?

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcTranslationOperation/DocEntity.xml
SergejMuhic: Well this is a coordinate, it depends on what it references. So, I am not too thrilled about changing it into the proposed form.

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcTranslationOperation/DocEntity.xml
SergejMuhic: Will change this in ifcdoc and re-serialize.

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcTranslationOperation/DocEntity.xml
SergejMuhic: Will also re-serialize

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcTranslationOperation/DocEntity.xml
SergejMuhic: re-serialize

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcTranslationOperation/DocEntity.xml
SergejMuhic: re-serialize

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcTranslationOperation/DocEntity.xml
SergejMuhic: Will re-serialize.

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcTranslationOperation/DocEntity.xml
pjanck: The description talks about `map coordinates`, then it would only be consistent to also talk about `geographic coordinates`.

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcTranslationOperation/Documentation.md
pjanck: I propose to use `rigid transformation` for this; need to think about the exact wording, though.

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcGeographicCRS/DocEntity.xml
pjanck: Yes, similarly to how `IfcProjectedCRS` has attributes for `MapZone` etc. this is the parallel in geographic CRS.

anborr (approved):

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcGeographicCRS/DocEntity.xml
```diff
@@ -0,0 +1,8 @@
...
+	</Attributes>
+</DocEntity>
```
pjanck, line 34: We need to put a WHERE rule to ensure the Unit is a PlaneAngleUnit.

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcGeographicCRS/Documentation.md
```diff
@@ -0,0 +1 @@
+A _IfcGeographicCRS_ is a CRS that uses a three-dimensional ellipsoid surface to determine locations on the Earth. Any location on Earth can be referenced by a point with longitude and latitude coordinates.
```
pjanck:
```suggestion
A _IfcGeographicCRS_ is a CRS that uses a three-dimensional ellipsoid surface to determine locations on the Earth. Any location on Earth can be referenced by a point with longitude and latitude coordinates and ellipsoidal height.
```

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcTranslationOperation/DocEntity.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="utf-8"?>
+<DocEntity xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" Name="IfcTranslationOperation" UniqueId="946a945b-8a4a-4e2c-9840-b56032727b5f" BaseDefinition="IfcCoordinateOperation" EntityFlags="32">
```
pjanck:
```suggestion
<DocEntity xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" Name="IfcRigidOperation" UniqueId="946a945b-8a4a-4e2c-9840-b56032727b5f" BaseDefinition="IfcCoordinateOperation" EntityFlags="32">
```

I believe this was the agreed name in the end.

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcTranslationOperation/DocEntity.xml
```diff
@@ -0,0 +1,33 @@
...
+	<Attributes>
+		<DocAttribute Name="FirstCoordinate" UniqueId="f2534a2f-e119-4a52-b1cf-a55c170eea45" DefinedType="IfcMeasureValue">
+			<Documentation>The first coordinate of the translation. Can be a length measure in case of map coordinates and angle in case of geographic reference systems.</Documentation>
```
pjanck:
```suggestion
			<Documentation>The first coordinate of the translation. Can be a length measure in case of map coordinates or a solid angle measure in case of geographic reference systems.</Documentation>
```

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcTranslationOperation/Documentation.md
```diff
@@ -0,0 +1 @@
+A translation operation specifies an offset in the coordinate reference system. It does not specify any conversion or distortion only a translation. The coodinate axes of the _IfcGeometricRepresentationContext_ can rotate and set a new basis coordinate system.
```
pjanck:
```suggestion
A rigid operation specifies an offset in the coordinate reference system. It does not specify any conversion or distortion only a translation. The coordinate axes of the _IfcGeometricRepresentationContext_ can rotate and set a new basis coordinate system.
```

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcWellKnownTextCRS/DocEntity.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<DocEntity xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" Name="IfcWellKnownTextCRS" UniqueId="2c5451db-1148-4295-9d0e-fd2a3abfe638" BaseDefinition="IfcCoordinateReferenceSystem" EntityFlags="32">
```
pjanck: I'd rather see a specialized WKT atribute on each `IfcCoordinateReferenceSystem`.

pjanck (changes requested):
Some suggestions. Regarding the WKT I propose a meeting.

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcWellKnownTextCRS/DocEntity.xml
ChristianClemen: I agree with PJanck. An [OGC/ISO standard](https://www.ogc.org/standards/wkt-crs) compliant string containing the CRS information in a computer interpretable way would be of great help for geospatial software. 

All necessary scale adjustments (e.g. x/y-scale based on the height (h) above the ellipsoid) can be managed in the WKT. In terms of BIM quality assurance, each IFC data delivery can be checked to see whether it actually shows the agreed WKT string.

It is clear to me that it is not typical of IFC to write so many "chunks" of information as a single attribute in a string. However, that would be a good step in the right direction for the goal of automation-compatible BIM/GIS integration

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcGeographicCRS/DocEntity.xml
```diff
@@ -2,7 +2,19 @@
...
+			<Definition>
+				<DocDefinitionRef xsi:nil="true" href="IfcIdentifier_1dG8O0DIbDBxufYev6dUDH" />
+			</Definition>
+		</DocAttribute>
 		<DocAttribute Name="Unit" UniqueId="82bf71dc-4f74-483b-84c7-70c84a5ce22d" DefinedType="IfcNamedUnit" AttributeFlags="1" />
 	</Attributes>
+	<WhereRules>
+		<DocWhereRule Name="UnitIsPlaneAngle" UniqueId="45472275-87bf-4f5e-a1ad-80be7f068991">
+			<Documentation>The type of Unit shall be _IfcPlaneAngleMeasure_.</Documentation>
+			<Expression>&apos;IFC4x3.IFCPLANEANGLEMEASURE&apos; IN TYPEOF(SELF.Unit)</Expression>
```
pjanck:
```suggestion
			<Expression>NOT(EXISTS(Unit)) OR (Unit.UnitType = IfcUnitEnum.PLANEANGLEUNIT)</Expression>
```

IFCPLANEANGLE*MEASURE* is probably false...

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcMapConversion/DocEntity.xml
```diff
@@ -42,5 +42,11 @@ _XAxisAbscissa_ it provides the direction of the local x axis within the horizon
 			<Documentation>Scale Z by which the length measures in local engineering CS have to be multiplied to get map coordinates</Documentation>
 		</DocAttribute>
 	</Attributes>
+	<WhereRules>
+		<DocWhereRule Name="NoGeographicCRS" UniqueId="ac9fb308-c18a-436f-8ec4-d0591f1dca67">
+			<Documentation>_IfcCoordinateOperation_.TargetCRS is not of type _IfcGeographicCRS_.</Documentation>
+			<Expression>NOT(&apos;IFC4x3.IFCGEOGRAPHICCRS&apos; IN TYPEOF(SELF\IfcCoordinateOperation.TargetCRS))</Expression>
```
pjanck: I agree with restricting the type, however: consider restricting *positively* to only allow for *IfcProjectedCRS*.

Proposal:
- Name: *OnlyProjectedCRS*
- `&apos;IFC4x3.IFCPROJECTEDCRS&apos; IN TYPEOF(SELF\IfcCoordinateOperation.TargetCRS)`

Rationale: 
- if, at any point in the future, there will be another subtype of *IfcCoordinateReferenceSystem*, there is no need to fix this rule
- it makes sense that only the projected CRS can be user with map projection

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcRigidOperation/DocEntity.xml
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
-<DocEntity xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" Name="IfcTranslationOperation" UniqueId="946a945b-8a4a-4e2c-9840-b56032727b5f" BaseDefinition="IfcCoordinateOperation" EntityFlags="32">
+<DocEntity xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" Name="IfcRigidOperation" UniqueId="946a945b-8a4a-4e2c-9840-b56032727b5f" BaseDefinition="IfcCoordinateOperation" EntityFlags="32">
 	<Attributes>
 		<DocAttribute Name="FirstCoordinate" UniqueId="f2534a2f-e119-4a52-b1cf-a55c170eea45" DefinedType="IfcMeasureValue">
-			<Documentation>The first coordinate of the translation. Can be a length measure in case of map coordinates and angle in case of geographic reference systems.</Documentation>
+			<Documentation>The first coordinate of the translation. Can be a length measure in case of map coordinates or a solid angle measure in case of geographic reference systems.</Documentation>
```
pjanck:
```suggestion
			<Documentation>The first coordinate of the translation. Can be a length measure in case of map coordinates or a plane angle measure in case of geographic reference systems.</Documentation>
```

Change as well with the second coordinate.

I can't comment below, but see lines 22 and 25 of this file: *IfcPositivePlaneAngleMeasure* should be *IfcPlaneAngleMeasure*.

pjanck (changes requested):

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcWellKnownText/DocEntity.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<DocEntity xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" id="IfcWellKnownText" Name="IfcWellKnownText" UniqueId="2c5451db-1148-4295-9d0e-fd2a3abfe638" BaseDefinition="IfcGeometricRepresentationItem" EntityFlags="32">
+	<Attributes>
+		<DocAttribute Name="WellKnownText" UniqueId="64c67b4d-f548-4d88-b1fb-b770f4f5bc58" DefinedType="IfcText">
+			<Documentation>Text string according to ISO/IEC 13249. If _CoordinateReferenceSystem_ is populated, the the representation is a coordinate reference system according to ISO 19162.</Documentation>
```
pjanck:
```suggestion
			<Documentation>Text string according to ISO/IEC 13249. If _CoordinateReferenceSystem_ is populated, then the representation is a coordinate reference system according to ISO 19162.</Documentation>
```

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcWellKnownText/DocEntity.xml
```diff
@@ -0,0 +1,15 @@
...
+		</DocAttribute>
+		<DocAttribute Name="CoordinateReferenceSystem" UniqueId="32a92b28-a350-4750-93a2-3b3207705c9f" DefinedType="IfcCoordinateReferenceSystem" AttributeFlags="1">
+			<Documentation>The _IfcCoordinateReferenceSystem_ that is represented by _WellKnownText_.</Documentation>
+			<Definition>
+				<DocEntity xsi:nil="true" href="IfcCoordinateReferenceSystem" />
+			</Definition>
```
pjanck, line 14: If I understand this correctly, this enables that a CRS can be defined with a WKT instead of an EPSG. 

Propose to make `IfcCoordinateReferenceSystem::Name` attribute optional, with a new WHERE rule ensuring that either `Name` or the newly added `HasWellKnowTextRepresentation` are existent.

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcWellKnownText/Documentation.md
```diff
@@ -0,0 +1 @@
+_IfcWellKnownText_ is a geometric representation item that is defined by a text string in _WellKnowText_ the format of which is specified in ISO 19162.
```
pjanck: The attribute `IfcWellKnownText::WellKnownText`'s documentation and this documentation are not on on par. While the former speaks also of ISO 13249, this one does not.

pjanck (changes requested):

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcWellKnownText/Documentation.md
pjanck: I see this has been fixed while reviewing :P (94965d3)

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcWellKnownTextCRS/DocEntity.xml
pjanck: This has now been addressed with a new `IfcWellKnownText` entity.

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/DocSchema.xml
SergejMuhic: Newly added get an id. Could change this but not priority.

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcCoordinateReferenceSystem/DocEntity.xml
```diff
@@ -44,6 +44,18 @@ ELLIPSOID[&amp;lt;name&amp;gt;,&amp;lt;major axis&amp;gt;,&amp;lt;inverse flatte
...
+		<DocAttribute Name="HasWellKnowTextRepresentation" UniqueId="9b9f3bbc-d2fc-4157-a1c7-9a86730b2575" DefinedType="IfcWellKnownText" AttributeFlags="1" Inverse="CoordinateReferenceSystem">
+			<Documentation>Well known text (string) according to ISO 19162.</Documentation>
+			<Definition>
+				<DocEntity xsi:nil="true" href="IfcWellKnownText" />
+			</Definition>
+		</DocAttribute>
 	</Attributes>
+	<WhereRules>
+		<DocWhereRule Name="CoordinateSystemSpecified" UniqueId="72c056ab-f8e2-4c95-80f4-3a41604c4d90">
+			<Documentation>Ensures that the coordinate reference system is properly specified by either a reference to EPSG in _Name: or well known text in _HasWellKnownTextRepresentation_.</Documentation>
```
pjanck:
```suggestion
			<Documentation>Ensures that the coordinate reference system is properly specified by either a reference to an EPSG code in _Name_ or a well known text in _HasWellKnownTextRepresentation_.</Documentation>
```

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcCoordinateReferenceSystem/DocEntity.xml
```diff
@@ -44,6 +44,18 @@ ELLIPSOID[&amp;lt;name&amp;gt;,&amp;lt;major axis&amp;gt;,&amp;lt;inverse flatte
 		<DocAttribute Name="HasCoordinateOperation" UniqueId="83795333-6df7-46ac-b61e-5101640f8615" DefinedType="IfcCoordinateOperation" AggregationType="3" AggregationLower="0" AggregationUpper="1" Inverse="SourceCRS" XsdFormat="attribute">
 			<Documentation>Indicates conversion between coordinate systems. In particular it refers to an _IfcCoordinateOperation_ between this coordinate reference system, and another Geographic coordinate reference system.</Documentation>
 		</DocAttribute>
+		<DocAttribute Name="HasWellKnowTextRepresentation" UniqueId="9b9f3bbc-d2fc-4157-a1c7-9a86730b2575" DefinedType="IfcWellKnownText" AttributeFlags="1" Inverse="CoordinateReferenceSystem">
+			<Documentation>Well known text (string) according to ISO 19162.</Documentation>
```
pjanck: I would add a note, saying that if EPSG code is given in *Name* as well, the *Name* takes precedence.

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcGeographicCRS/DocEntity.xml
```diff
@@ -0,0 +1,20 @@
...
+			<Definition>
+				<DocDefinitionRef xsi:nil="true" href="IfcIdentifier_1dG8O0DIbDBxufYev6dUDH" />
+			</Definition>
+		</DocAttribute>
+		<DocAttribute Name="Unit" UniqueId="82bf71dc-4f74-483b-84c7-70c84a5ce22d" DefinedType="IfcNamedUnit" AttributeFlags="1" />
+	</Attributes>
+	<WhereRules>
+		<DocWhereRule Name="UnitIsPlaneAngle" UniqueId="45472275-87bf-4f5e-a1ad-80be7f068991">
+			<Documentation>The type of Unit shall be _IfcPlaneAngleMeasure_.</Documentation>
+			<Expression>SELF.Unit = IfcUnitEnum.PLANEANGLE</Expression>
```
pjanck:
```suggestion
			<Expression>NOT(EXISTS(SELF.Unit)) || SELF.Unit.UnitType = IfcUnitEnum.PLANEANGLE</Expression>
```

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcGeographicCRS/DocEntity.xml
```diff
@@ -0,0 +1,20 @@
...
+<DocEntity xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" Name="IfcGeographicCRS" UniqueId="2d6be663-8d5e-42cc-b5b4-13bb37f41262" BaseDefinition="IfcCoordinateReferenceSystem" EntityFlags="32">
+	<Attributes>
+		<DocAttribute Name="GeodeticDatum" UniqueId="b4d581f7-612a-4e98-963a-95ad06714397" DefinedType="IfcIdentifier" AttributeFlags="1" />
+		<DocAttribute Name="PrimeMeridian" UniqueId="b2c9cf50-b4bb-4979-92d8-c44a05f22954" DefinedType="IfcIdentifier" AttributeFlags="1">
+			<Documentation>The identification of the meridian defining zero longitude in the used geographic CRS.</Documentation>
+			<Definition>
+				<DocDefinitionRef xsi:nil="true" href="IfcIdentifier_1dG8O0DIbDBxufYev6dUDH" />
+			</Definition>
+		</DocAttribute>
+		<DocAttribute Name="Unit" UniqueId="82bf71dc-4f74-483b-84c7-70c84a5ce22d" DefinedType="IfcNamedUnit" AttributeFlags="1" />
```
pjanck: Have been thinking about the *height above the ellipsoid* - would it require a separate unit as well? If yes, another WHERE rule ensuring it is a Length Unit would also be needed.

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcGeographicCRS/Documentation.md
```diff
@@ -0,0 +1 @@
+A _IfcGeographicCRS_ is a CRS that uses a three-dimensional ellipsoid surface to determine locations on the Earth. Any location on Earth can be referenced by a point with longitude and latitude coordinates and the height above the ellipsoid.
```
pjanck:
```suggestion
A _IfcGeographicCRS_ is a CRS that uses a three-dimensional ellipsoid surface to determine locations on the Earth. Any location on Earth can be referenced by a point with longitude and latitude coordinates and the height above/below the ellipsoid.
```

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcMapConversion/DocEntity.xml
```diff
@@ -42,5 +42,11 @@ _XAxisAbscissa_ it provides the direction of the local x axis within the horizon
 			<Documentation>Scale Z by which the length measures in local engineering CS have to be multiplied to get map coordinates</Documentation>
 		</DocAttribute>
 	</Attributes>
+	<WhereRules>
+		<DocWhereRule Name="OnlyProjectedCRS" UniqueId="ac9fb308-c18a-436f-8ec4-d0591f1dca67">
+			<Documentation>_IfcCoordinateOperation_.TargetCRS is not of type _IfcGeographicCRS_.</Documentation>
```
pjanck:
```suggestion
			<Documentation>_IfcCoordinateOperation_.TargetCRS is of type _IfcProjectedCRS_.</Documentation>
```

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcRigidOperation/DocEntity.xml
```diff
@@ -0,0 +1,33 @@
...
+<DocEntity xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" Name="IfcRigidOperation" UniqueId="946a945b-8a4a-4e2c-9840-b56032727b5f" BaseDefinition="IfcCoordinateOperation" EntityFlags="32">
+	<Attributes>
+		<DocAttribute Name="FirstCoordinate" UniqueId="f2534a2f-e119-4a52-b1cf-a55c170eea45" DefinedType="IfcMeasureValue">
+			<Documentation>The first coordinate of the translation. Can be a length measure in case of map coordinates or a plane angle measure in case of geographic reference systems.</Documentation>
+			<Definition>
+				<DocDefinitionRef xsi:nil="true" href="IfcMeasureValue_0zhETP7Hz2E94mtLjvV20s" />
+			</Definition>
+		</DocAttribute>
+		<DocAttribute Name="SecondCoordinate" UniqueId="366b1fda-0ecf-4891-abb5-212d322b50fb" DefinedType="IfcMeasureValue">
+			<Documentation>The second coordinate of translation. Can be a length measure in case of map coordinates and plane angle in case of geographic reference systems.</Documentation>
```
pjanck:
```suggestion
			<Documentation>The second coordinate of the translation. Can be a length measure in case of map coordinates or a plane angle measure in case of geographic reference systems.</Documentation>
```

Making it equal to the FirstCoordinate documentation.

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcRigidOperation/DocEntity.xml
```diff
@@ -0,0 +1,33 @@
...
+				<DocDefinitionRef xsi:nil="true" href="IfcMeasureValue_0zhETP7Hz2E94mtLjvV20s" />
+			</Definition>
+		</DocAttribute>
+		<DocAttribute Name="Height" UniqueId="caad8793-db34-4377-b751-b1591d8f3fe3" DefinedType="IfcLengthMeasure" AttributeFlags="1">
+			<Documentation>Height above (positive) or below (negative) the coordinate surface.</Documentation>
+		</DocAttribute>
+	</Attributes>
+	<WhereRules>
+		<DocWhereRule Name="CoordinateType" UniqueId="e9bb63a8-0676-42b5-93db-fa2803a66ae2">
+			<Documentation>Restricts the value type of _FirstCoordinate_ and _SecondCoordinate_ to _IfcLengthMeasure_ and _IfcPositivePlaneAngleMeasure_.</Documentation>
```
pjanck:
```suggestion
			<Documentation>Restricts the value type of _FirstCoordinate_ and _SecondCoordinate_ to _IfcLengthMeasure_ and _IfcPlaneAngleMeasure_.</Documentation>
```

Is there a specific reason to use *positive* restriction?

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcRigidOperation/DocEntity.xml
```diff
@@ -0,0 +1,33 @@
...
+			<Expression>(&apos;IFCMEASURERESOURCE.IFCLENGTHMEASURE&apos; IN TYPEOF(FirstCoordinate)
+OR
+&apos;IFCMEASURERESOURCE.IFCPOSITIVEPLANEANGLEMEASURE&apos; IN TYPEOF(FirstCoordinate))
```
pjanck:
```suggestion
&apos;IFCMEASURERESOURCE.IFCPLANEANGLEMEASURE&apos; IN TYPEOF(FirstCoordinate))
```

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcRigidOperation/DocEntity.xml
```diff
@@ -0,0 +1,33 @@
...
+AND
+(&apos;IFCMEASURERESOURCE.IFCLENGTHMEASURE&apos; IN TYPEOF(SecondCoordinate)
+OR
+&apos;IFCMEASURERESOURCE.IFCPOSITIVEPLANEANGLEMEASURE&apos; IN TYPEOF(SecondCoordinate))</Expression>
```
pjanck:
```suggestion
&apos;IFCMEASURERESOURCE.IFCPLANEANGLEMEASURE&apos; IN TYPEOF(SecondCoordinate))</Expression>
```

pjanck (changes requested):
Looked thoroughly through all the changes of the PR from the start. Some suggestions and proposals. Happy to discuss!

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcGeographicCRS/DocEntity.xml
SergejMuhic: There is no height. Should there be? If I add it, I will make it an IfcLengthMeasure so the project units would apply.

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcCoordinateReferenceSystem/DocEntity.xml
```diff
@@ -9,7 +9,7 @@
 		<DocSubtype UniqueId="065cabab-0e6e-4c55-92fa-b7a1fcf53185" DefinedType="IfcProjectedCRS" />
```
larswik, line 9: Should IfcGeographicCRS be added or just remove this entirely?

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcCoordinateReferenceSystem/DocEntity.xml
larswik: Agree.

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcGeographicCRS/DocEntity.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<DocEntity xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" Name="IfcGeographicCRS" UniqueId="2d6be663-8d5e-42cc-b5b4-13bb37f41262" BaseDefinition="IfcCoordinateReferenceSystem" EntityFlags="32">
+	<Attributes>
+		<DocAttribute Name="GeodeticDatum" UniqueId="b4d581f7-612a-4e98-963a-95ad06714397" DefinedType="IfcIdentifier" AttributeFlags="1" />
```
larswik: GeodeticDatum also in IfcCoordinateReferenceSystem. Is this by design?

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcGeographicCRS/DocEntity.xml
larswik: I read this as referencing the documentation text: "Any location on Earth can be referenced by a point with longitude and latitude coordinates and the **height above the ellipsoid**". I would guess that no action needed (same unit for all coordinates).

SergejMuhic: The problem @pjanck is pointing out is that `First` and `SecondCoordinate` can be either `IfcLenghtMeasure` or `IfcPlaneAngleMeasure`. However, Height is already restricted so no need to do anything there.

Please look at `IfcRigidOperation`. The type there is still `IfcPositivePlaneAngleMeasure`.

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcRigidOperation/DocEntity.xml
SergejMuhic: I don't know. It is like that for a long time now. I don't even know if anyone demanded it. Would we consider IfcCompoundPlaneAngleMeasure even?

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcGeographicCRS/DocEntity.xml
SergejMuhic: Good question. @pjanck why do we have two GeodertiDatums?

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcRigidOperation/DocEntity.xml
pjanck: Good suggestion. Therefore, allowed would be:
- `IfcLengthMeasure`
- `IfcPlaneAngleMeasure`
- `IfcCompoundPlaneAngleMeasure`

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcGeographicCRS/DocEntity.xml
pjanck: > The problem @pjanck is pointing out is that `First` and `SecondCoordinate` can be either `IfcLenghtMeasure` or `IfcPlaneAngleMeasure`. However, Height is already restricted so no need to do anything there.

I wan't refereing to that. Rather, I was addressing

> the project units would apply

I would like to *allow* (similarly to how `IfcProjectedCRS` defines a Unit, or how for First and SecondCoordinate there is a Unit) for a different height-unit of the `IfcGeographicCRS`, compared to the project units. This way, a project using feet could still be located x meters above/below an ellipsoid (using `IfcRigidOperation`).

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcGeographicCRS/DocEntity.xml
pjanck: Right, I see. There is a change missing: `GeodeticDatum` and `VerticalDatum` shall be demoted from `IfcCoordinateReferenceSystem` to the sub-entity `IfcProjectedCRS`. Rationale: they are projected CRS specifics and not necessary a part of *any* CRS (see ISO 19111).

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcGeographicCRS/DocEntity.xml
pjanck: I don't see this change in the commits, although it is resolved. Care to double check?

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcMapConversion/DocEntity.xml
```diff
@@ -44,7 +44,7 @@ _XAxisAbscissa_ it provides the direction of the local x axis within the horizon
 	</Attributes>
 	<WhereRules>
 		<DocWhereRule Name="OnlyProjectedCRS" UniqueId="ac9fb308-c18a-436f-8ec4-d0591f1dca67">
-			<Documentation>_IfcCoordinateOperation_.TargetCRS is not of type _IfcGeographicCRS_.</Documentation>
+			<Documentation>_IfcCoordinateOperation_.TargetCRS is of type _IfcGeographicCRS_.</Documentation>
```
pjanck, line 41:
```suggestion
			<Documentation>_IfcCoordinateOperation_.TargetCRS is of type _IfcProjectedCRS_.</Documentation>
```

pjanck (changes requested):

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcGeographicCRS/DocEntity.xml
```diff
@@ -0,0 +1,19 @@
...
+			<Documentation>The identification of the meridian defining zero longitude in the used geographic CRS.</Documentation>
+			<Definition>
+				<DocDefinitionRef xsi:nil="true" href="IfcIdentifier_1dG8O0DIbDBxufYev6dUDH" />
+			</Definition>
+		</DocAttribute>
+		<DocAttribute Name="Unit" UniqueId="82bf71dc-4f74-483b-84c7-70c84a5ce22d" DefinedType="IfcNamedUnit" AttributeFlags="1" />
+	</Attributes>
+	<WhereRules>
+		<DocWhereRule Name="UnitIsPlaneAngle" UniqueId="45472275-87bf-4f5e-a1ad-80be7f068991">
+			<Documentation>The type of Unit in the operation shall be _IfcPlaneAngleMeasure_.</Documentation>
```
pjanck: I'm confused a bit, when a *unit* shall be *measure*. Perhaps we can change the wording to better reflect that the unit should be a plane angle *unit*? (I'm unsure how this is worded in the rest of the standard.)

pjanck (changes requested):
There are also some left-overs from previous reviews. Happy to discuss further! 

I really like the solution of WKT with the inverses. <3

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcGeographicCRS/DocEntity.xml
```diff
@@ -0,0 +1,35 @@
...
+	<Attributes>
+		<DocAttribute Name="PrimeMeridian" UniqueId="b2c9cf50-b4bb-4979-92d8-c44a05f22954" DefinedType="IfcIdentifier" AttributeFlags="1">
+			<Documentation>The identification of the meridian defining zero longitude in the used geographic CRS.</Documentation>
+			<Definition>
+				<DocDefinitionRef xsi:nil="true" href="IfcIdentifier_1dG8O0DIbDBxufYev6dUDH" />
+			</Definition>
+		</DocAttribute>
+		<DocAttribute Name="AngleUnit" UniqueId="82bf71dc-4f74-483b-84c7-70c84a5ce22d" DefinedType="IfcNamedUnit" AttributeFlags="1">
+			<Documentation>Unit of latitude and longitude coordinate axes composing the geographic coordinate system.
+&gt; NOTE 1&amp;nbsp; Only plane angle measures are in scope and both longitude and latitude coordinate axes of the geogrpahic coordinate system shall have the same plane angle unit.
```
pjanck:
```suggestion
&gt; NOTE 1&amp;nbsp; Only plane angle measures are in scope and both longitude and latitude coordinate axes of the geographic coordinate system shall have the same plane angle unit.
```

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcGeographicCRS/DocEntity.xml
```diff
@@ -0,0 +1,35 @@
...
+&gt; NOTE 2&amp;nbsp; If AngleUnit is omitted, the unit for latitude and longitude coordiante axes is taken from the default project angle units, as stated in IfcProject.UnitInContext.</Documentation>
```
pjanck:
```suggestion
&gt; NOTE 2&amp;nbsp; If AngleUnit is omitted, the unit for latitude and longitude coordinate axes is taken from the default project angle units, as stated in IfcProject.UnitInContext.</Documentation>
```

pjanck (changes requested):
Two typos.

pjanck:
I've compiled an EXPRESS-G representation of changes to be presented next week at the 3D GeoInfo conference in Sydney, Australia. I hope that I have captured all changes and am providing this to the community for review and discussion.

grafik

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcRepresentationResource/Entities/IfcMapConversion/DocEntity.xml
SergejMuhic: Change merged.

PR #351, opened 2022-02-14
Update IfcReferent

### LINKED ISSUE #344: Removal of attribute IfcReferent.RestartDistance
As title says, we want to remove `IfcReferent.RestartDistance` in favour of the `Pset_Stationing` property set.

larswik:
Deprecate RestartDistance.
Fixes #344


clfey:
CLFEY : No opinion concerning removal or deprecation.

czapplitec:
I would actually prefer to remove. There are some changes being done on the schema in parallel by bSI. We don't have to keep the "no-change-on-schema" idea.

larswik:
If everybody agrees on a remove, I will do that. I assume that the same goes for MILEPOST and KILOPOST as well.

clfey (approved):

SergejMuhic:
> I would actually prefer to remove. There are some changes being done on the schema in parallel by bSI. We don't have to keep the "no-change-on-schema" idea.

Me too. The resolution was to remove. Since it was added in 4.1 which has been withdrawn, I would say that it is safe to remove.

larswik:
RestartDistance removed now.

czapplitec (approved):

SergejMuhic (approved):

PR #352, opened 2022-02-14
IfcReferentTypeEnum constants updated

### LINKED ISSUE #345: Improve documentation for IfcReferentTypeEnum STATION and POSITION
As agreed in the stationing meetings we are keeping both predefined types and improving their documentation.

larswik:
Attempted to fix #345.

### DIFF
--- a/IFC4x3/Constants/s/STATION_3mZYci1kzBIfEoa0zx_hua.xml
+++ b/IFC4x3/Constants/s/STATION_3mZYci1kzBIfEoa0zx_hua.xml
@@ -1,3 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<DocConstant xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" id="STATION_3mZYci1kzBIfEoa0zx_hua" Name="STATION" UniqueId="f08e29ac-06ef-4b4a-93b2-900f7bfabe24" Status="Deprecated" />
+<DocConstant xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" id="STATION_3mZYci1kzBIfEoa0zx_hua" Name="STATION" UniqueId="f08e29ac-06ef-4b4a-93b2-900f7bfabe24">
+	<Documentation>Used to represent a linear location given by the linear element being measured (the _IfcAlignment_ into which the _IfcReferent_ is nested), a station value or a station equation using _Pset_Stationing_ and optionally the method of measurement (_Pset_LinearReferencingMethod_). If a linear referencing method is specified for the station, it overrides any linear referencing method specified for the alignment.</Documentation>
+</DocConstant>
 
--- a/IFC4x3/Constants/s/STATION_3mZYci1kzBIfEoa0zx_hua.xml
+++ b/IFC4x3/Constants/s/STATION_3mZYci1kzBIfEoa0zx_hua.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <DocConstant xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" id="STATION_3mZYci1kzBIfEoa0zx_hua" Name="STATION" UniqueId="f08e29ac-06ef-4b4a-93b2-900f7bfabe24">
-	<Documentation>Used to represent a linear location given by the linear element being measured (the _IfcAlignment_ into which the _IfcReferent_ is nested), a station value or a station equation using _Pset_Stationing_ and optionally the method of measurement (_Pset_LinearReferencingMethod_). If a linear referencing method is specified for the station, it overrides any linear referencing method specified for the alignment.</Documentation>
+	<Documentation>Represents a linearly measured location along the linear positioning element into which the _IfcReferent_ is nested. The location is described by a station value or a station equation using _Pset_Stationing_ and optionally the method of measurement (_Pset_LinearReferencingMethod_). If a linear referencing method is specified for the _STATION_ referent, it then overrides any linear referencing method specified for the alignment.</Documentation>
 </DocConstant>
 
--- a/IFC4x3/Constants/s/STATION_3mZYci1kzBIfEoa0zx_hua.xml
+++ b/IFC4x3/Constants/s/STATION_3mZYci1kzBIfEoa0zx_hua.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <DocConstant xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" id="STATION_3mZYci1kzBIfEoa0zx_hua" Name="STATION" UniqueId="f08e29ac-06ef-4b4a-93b2-900f7bfabe24">
-	<Documentation>Represents a linearly measured location along the linear positioning element into which the _IfcReferent_ is nested. The location is described by a station value or a station equation using _Pset_Stationing_ and optionally the method of measurement (_Pset_LinearReferencingMethod_). If a linear referencing method is specified for the _STATION_ referent, it then overrides any linear referencing method specified for the alignment.</Documentation>
+	<Documentation>Represents a linearly measured location along the linear positioning element into which the _IfcReferent_ is nested. The location is described by a station value or a station equation using _Pset_Stationing_ and optionally the method of measurement (_Pset_LinearReferencingMethod_). If a linear referencing method is specified for the STATION referent, it overrides any linear referencing method specified for the alignment.</Documentation>
 </DocConstant>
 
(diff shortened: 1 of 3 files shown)


clfey:
CLFEY: My opinion was to remove POSITION. How is POSITION useful / needed?

larswik:
My interpretation of the decision/issue was to keep both and that we need to clearly distinguish the two with good definitions. I have made a first attempt in this pull request and expect review/comments/suggestions. If we can't make a clear enough distinction, I agree that we should remove POSITION.

clfey (commented):
It is suggested to rethink if POSITION is really needed as a PDT, since we have STATION.

- All referents may possess a PointByDistanceExpression, which probably is a "position".
- The LRMName and LRMType Pset attributes allows users to shape a STATION referent into any possible use he/she can think of.

czapplitec:
I am neutral for this one, keep it or not. My understanding is that it is a generic position concept. It could be used in:
- a position that has lateral, vertical or longitudinal offset based on DistanceAlong (I guess we don't expect STATION and REFRENCEMARKER to have these)
- or any generic position that may have nothing to do with linear assets. However, this scenario does not comply to the current definition of IfcReferent, so I may misunderstand.

clfey:
My opinion is that all referents may possess a PointByDistanceExpression, as long as the alignment that it has been nested to has horizontal geometry. This information can then be picked up by an authoring tool. The authoring tool will produce a STATION referent which may have Pset_Stationing.Station as its numerical value. This referent has an IfcReferent.Name that contains any information that this authoring tool wants to present, as a USERDEFINED LRMType.

In one application, this is exemplified as comprehensive information shown on-screen as you move the cursor in the authoring tool. This information may then be packed into a STATION referent with LRMType "USERDEFINED" and stored in an Ifc file at the time of export to IFC:

image

The text to be stored as the IfcReferent.Name attribute could in this case be "[Connection A]@462,344687 / Elevation 9,045 masl / Lateral Offset -4.000 m".  No need for a POSITION referent - all info is contained in the STATION.

larswik:
I would like some input from @SergejMuhic on this one. My understanding is similar to @czapplitec. I tried to discriminate POSITION from the rest by stating that a STATION or REFERENCEMARKER are always linearly located while a POSITION can represent any arbitrary location, e.g. by using local placement.

clfey:
...meaning that a POSITION does not have a POintByDistanceExpression (no DistanceAlong), but could have just a DirectPosition coordinates?

In one project, we just received a file containing 1000 referents. Each referent (one row in an Excel spreadsheet) has a name such as "18.0", "18.1", "18.2" ...."100.0", every ca 100 meters. The geometry of the tracks is still unknown (track survey is still ongoing), but each of these "kilometration board referents" were stated with X Y Z coordinates. Are they not REFERENCEMARKERs? 

Please give one or two use cases for POSITION, in a railway project context.

larswik:
Removed KILOPOINT and MILEPOINT. POSITION remains.

larswik:
I see IfcReferent as a generic (and not railway specific) concept. Is it useful then to separate between stationing along an alignment and other “known and named locations”?

czapplitec:
If we agree that STATION could have offset too (refers to Claus’ image in the PR), then I don’t see use case for POSITION in linear infrastructure projects, unless we just want some generic POSITION concept. This means we could either remove it, or keep it there but don’t need to use it in railway.
@SergejMuhic, could you give your thoughts about POSITION, and if keep it, make the documentation?

SergejMuhic:
I cannot foresee use cases but if we are introducing a semantic representation of a point based location it only seems reasonable to have the `.POSITION.` option in the specification. In the case where there is no connection to an alignment and therefore no linear referencing conceptualization of a position a predefined type with the meaning of a position makes sense in my view. There are positions other than alignment based in our domains.

I do not see a reason to argue about this. I do not see it hurting anyone if properly documented. But, I would like to avoid specific concepts only applicable to a certain use case.

So, if @clfey feels so passionately about removing it, remove it. It could be used for modelling any known location that could be referenced by building elements. I am quite emotionless but I do see a place for it.

IFC4x3/Constants/s/STATION_3mZYci1kzBIfEoa0zx_hua.xml
```diff
@@ -1,3 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<DocConstant xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" id="STATION_3mZYci1kzBIfEoa0zx_hua" Name="STATION" UniqueId="f08e29ac-06ef-4b4a-93b2-900f7bfabe24" Status="Deprecated" />
+<DocConstant xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" id="STATION_3mZYci1kzBIfEoa0zx_hua" Name="STATION" UniqueId="f08e29ac-06ef-4b4a-93b2-900f7bfabe24">
+	<Documentation>Used to represent a linear location given by the linear element being measured (the _IfcAlignment_ into which the _IfcReferent_ is nested), a station value or a station equation using _Pset_Stationing_ and optionally the method of measurement (_Pset_LinearReferencingMethod_). If a linear referencing method is specified for the station, it overrides any linear referencing method specified for the alignment.</Documentation>
```
SergejMuhic: I am a bit skeptical referencing the psets in the description but since I do not see anyone else objecting, I guess it is fine.

Somehow the first sentence seems a bit confusing. Here a proposal:
```suggestion
	<Documentation>Represents a position linearly measured along the linear positioning element into which the _IfcReferent_ is nested. The position is described by a station value or a station equation using _Pset_Stationing_ and optionally the method of measurement (_Pset_LinearReferencingMethod_). If a linear referencing method is specified for the station, it overrides any linear referencing method specified for the alignment.</Documentation>
```

SergejMuhic (changes requested):
Opening a discussion on the documentation of `STATION`. 😄

IFC4x3/Constants/s/STATION_3mZYci1kzBIfEoa0zx_hua.xml
clfey: 1)  Concerning `.POSITION.` : Sergej writes "It could be used for modelling any known location that could be referenced by building elements.". I am fine with this, I just wanted to see a plausiable argument for a situation where it might be used. I withdraw the proposal to remove it.

2) Concerning the proposed text change for the .STATION. referent: Would it be posible to use the word "location" instead of "position", so we do not blend in the `.POSITION.`here? If so, the text would read as follows: 

`[STATION] Represents a linearly measured location  along the linear positioning element into which the _IfcReferent_ is nested. The location is described by a station value or a station equation using _Pset_Stationing_ and optionally the method of measurement (_Pset_LinearReferencingMethod_). If a linear referencing method is specified for the _STATION_ referent, ithen t overrides any linear referencing method specified for the alignment.`

SergejMuhic: I have propose position because my impression was that location is more physical in nature than position which can also be an abstract description. I am, however, no expert on this and am completely fine to go with the experts.

larswik: I updated according to proposal of @clfey above. If also .POSITION. doc is ok, I think that it is time for a merge.

pjanck (approved):
lftm

IFC4x3/Constants/s/STATION_3mZYci1kzBIfEoa0zx_hua.xml
```diff
@@ -1,3 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<DocConstant xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" id="STATION_3mZYci1kzBIfEoa0zx_hua" Name="STATION" UniqueId="f08e29ac-06ef-4b4a-93b2-900f7bfabe24" Status="Deprecated" />
+<DocConstant xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" id="STATION_3mZYci1kzBIfEoa0zx_hua" Name="STATION" UniqueId="f08e29ac-06ef-4b4a-93b2-900f7bfabe24">
+	<Documentation>Represents a linearly measured location along the linear positioning element into which the _IfcReferent_ is nested. The location is described by a station value or a station equation using _Pset_Stationing_ and optionally the method of measurement (_Pset_LinearReferencingMethod_). If a linear referencing method is specified for the _STATION_ referent, it then overrides any linear referencing method specified for the alignment.</Documentation>
```
SergejMuhic:
```suggestion
	<Documentation>Represents a linearly measured location along the linear positioning element into which the _IfcReferent_ is nested. The location is described by a station value or a station equation using _Pset_Stationing_ and optionally the method of measurement (_Pset_LinearReferencingMethod_). If a linear referencing method is specified for the STATION referent, it overrides any linear referencing method specified for the alignment.</Documentation>
```

IFC4x3/Constants/s/STATION_3mZYci1kzBIfEoa0zx_hua.xml
SergejMuhic: Made a slight adjustment to formatting and removed "then" from the last sentence.

SergejMuhic (approved):

SergejMuhic:
Looks good.

PR #354, opened 2022-02-14
Pset_LinearReferencingMethod applicability

### LINKED ISSUE #346: Pset_LinearReferencingMethod ApplicableType change to IfcReferent
Applicability shall be extended to `IfcReferent` for the possibility of overriding the one set on `IfcAlignment`.

larswik:
Fixes #346


czapplitec (approved):

clfey (commented):
I just realised that Pset_LinearReferencingMethod is applicable to IfcAlignment. I do not see how this is meant to be used. I understand why each referent needs to know its LRM. One particular alignment may be used as the nesting element for many referents, where some referents may for instance use LRM type "1" and the others use for instance LRM type "2". The alignment will be used as the Linear Element (as per the ISO interpretation) for nesting of both type "1" and type "2" referents - but the alignment can only mention an LRM Pset once - so which of the two LRMs in use should be mentioned? A Pset cannot be present twice in the alignment. 

**Proposed solution** Remove "IfcAlignment" from the list of applicable elements for PsetLinearReferencingMethod.

czapplitec:
Every Pset is optional for the applicable entitites. I interpret it as IfcAlignment could use Pset_LinearReferencingMethod and all the nested IfcReferents will inherit the information. If there are more than one LRM types in the nested IfcReferents, then:
- Either define LRM type for each IfcReferent, so each IfcReferent has a Pset_LinearReferencingMethod
- Or define a "main" LRM type for IfcAlignment, and the nested IfcReferents which have LRM type other than the "main" one could have Pset_LinearReferencingMethod to overwrite.

larswik:
I share the view of @czapplitec above.

clfey:
I see - then it makes senseto to apply the Pset to the nesting IfcAlignment as well.

larswik:
@SergejMuhic : This seems fine for merging now.

SergejMuhic (approved):

PR #417, opened 2022-05-10
Adding attribute `IfcDerivedUnit.Name` and enums `FILLER` and `FLASHING` to `IfcDiscreteAccessoryTypeEnum`

SergejMuhic:
1. Added `IfcDerivedUnit.Name` attribute and documentation
2. Added IfcDiscreteAccessoryTypeEnum FILLER and FLASHING and documentation

### DIFF
--- a/IFC4x3/Constants/f/FILLER_2yUKyb8RD0ZwgSrzMvR8e0.xml
+++ b/IFC4x3/Constants/f/FILLER_2yUKyb8RD0ZwgSrzMvR8e0.xml
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<DocConstant xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" id="FILLER_2yUKyb8RD0ZwgSrzMvR8e0" Name="FILLER" UniqueId="bc794f25-21b3-408f-aa9c-d7d5b96c8a00">
+	<Documentation>Sealant, gap filler rod, packing material or other used to close a gap.</Documentation>
+</DocConstant>
+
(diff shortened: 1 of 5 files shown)


pjanck (approved):
Lftm

IFC4x3/Constants/f/FILLER_2yUKyb8RD0ZwgSrzMvR8e0.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<DocConstant xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" id="FILLER_2yUKyb8RD0ZwgSrzMvR8e0" Name="FILLER" UniqueId="bc794f25-21b3-408f-aa9c-d7d5b96c8a00">
+	<Documentation>Sealant, gap filler rod, packing material or other used to close a gap.</Documentation>
+</DocConstant>
+
```
larswik, line 5: A question for IFC Tunnel rather than a comment: In IFC Tunnel, we defined the new entity IfcFillElement with a couple of PDTs. The entity currently has the brief definition "An element with the purpose of filling gaps between other elements.". There seems to be a slight overlap between IfcDiscreteAccessory/FILLER and IfcFillElement. Is this a concern?

larswik (commented):
Besides the FILLER-question lftm.

aothms (approved):

IFC4x3/Constants/f/FILLER_2yUKyb8RD0ZwgSrzMvR8e0.xml
SergejMuhic: I guess we can look into this but for now I would merge to have the same baseline.

PR #440, opened 2022-06-16
Add IfcProductData

### LINKED ISSUE #439: [IFC Tunnel] Add entity IfcProductData
Separated from issue #391.
IfcProductData shall be an abstract entity inherited from IfcProduct and will be the basis for both observations and voxel data (and maybe more things in the future).
Proposed definition: A kind of product with the purpose of providing additional information to other products such as observations. Product data can carry its own representation and shall be related to the main product using _IfcRelAssignsToProduct_.

larswik:
In schema IfcProductExtension.
Fixes #439 .

### DIFF
--- a/IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcComplementaryData/Documentation.md
+++ b/IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcComplementaryData/Documentation.md
@@ -1,0 +1,1 @@
+A kind of product with the purpose of providing additional information, such as observations,  to other products. Complementary data may carry its own representation and shall be related to the main product using _IfcRelAssignsToProduct_.
--- a/IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcComplementaryData/Documentation.md
+++ b/IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcComplementaryData/Documentation.md
@@ -1,1 +1,1 @@
-A kind of product with the purpose of providing additional information, such as observations,  to other products. Complementary data may carry its own representation and shall be related to the main product using _IfcRelAssignsToProduct_.
+A kind of product with the purpose of providing additional raw data, such as observations,  to other products. Complementary data may carry its own representation and shall be related to the main product using _IfcRelAssignsToProduct_.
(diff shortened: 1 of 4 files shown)


anborr (changes requested):
After checking the other subtypes of IfcProduct, I would prefer IfcDataElement.

SergejMuhic:
> After checking the other subtypes of IfcProduct, I would prefer IfcDataElement.

IfcData maybe. Having difficulties wrapping my head around what a data element would be.

larswik:
> > After checking the other subtypes of IfcProduct, I would prefer IfcDataElement.
> 
> IfcData maybe. Having difficulties wrapping my head around what a data element would be.

I can see that IfcProductData may lead the "product data templates" people in the wrong direction. However, "IfcData" feels very vague and generic, as does "IfcDataElement". What we want to achieve is an element that, besides having its own properties and possibly representations, should always be associated with another IfcProduct. Also, in IFC, a product is not restricted to items sold by manufacturers. One alternative, if we can't agree on a generic term, would be to have IfcObservation as a separate thing directly under IfcProduct and make IfcVoxelData a separate, rooted or non-rooted, thing which may mean that IfcVoxelData would again need a separate relationship to the product it describes. My take is that "IfcProductData" is the best alternative yet (it is "just" a term which no end-user will ever have to see). How about "IfcProductRelatedData"? "My kingdom for a term"!

SergejMuhic:
What else to add. When I proposed the name it seemed a natural extension of the IfcProduct entity. I never felt any connection to anything other than that. Anything beyond, is not my domain. I am therefore completely fine with IfcProductData.

anborr:
I am really afraid that we see ourselves run into unnecessary trouble when we use "Product Data" as most people in AEC will associate this with Manufacturer Sheets etc. 
Also, from a pure "terminology science" point of view, the more abstract term must always come at the end of the more precise term: "Car" -> "RacingCar". This would result in "IfcDataProduct" which might be fine. At the same time, some subtypes of `IfcProduct` have the ending "..element", such as `IfcSpatialElement`, `IfcLinearElement` and `IfcPositioningElement`, so `IfcDataElement` seemed to be a natural candidate... In the end, the thing we describe here is an element of the model, right? 
But we also have "..Item" in `IfcStructuralItem`. So what about `IfcDataItem`?

larswik:
IfcComplementaryData

larswik:
Changed name IfcProductData=>IfcComplementaryData. Great if this PR could be handled quickly since there are other dependent PR waiting to happen (IfcVoxelData & IfcGeoScienceObservation). Ping @SergejMuhic & @anborr .

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcComplementaryData/Documentation.md
```diff
@@ -0,0 +1 @@
+A kind of product with the purpose of providing additional information, such as observations,  to other products. Complementary data may carry its own representation and shall be related to the main product using _IfcRelAssignsToProduct_.
```
SergejMuhic:
```suggestion
A kind of product with the purpose of providing additional raw data, such as observations,  to other products. Complementary data may carry its own representation and shall be related to the main product using _IfcRelAssignsToProduct_.
```

SergejMuhic (approved):

PR #465, opened 2022-07-26
Removing IfcProxy and reference ObjectTypeEnum

SergejMuhic:
1. Removed `IfcProxy`
2. Removed `IfcObjectTypeEnum`

I would propose removing `IfcObjectTypeEnum` in the bSI documentation also.


TLiebich:
@aothms - should we follow suit? I.e. removing the obsolete and long deprecated _IfcObjectTypeEnum_. It is also used by _IfcRelAssign_ but again deprecated - deprecation note says:  _The attribute is deprecated and shall no longer be used. A NIL value should always be assigned._ - so the sementically more correct type would be IfcStrippedOptional (invented for exactly this purpose - to mark a removed attribute while maintaining the attribute order.

Then also the function _IfcCorrectObjectAssignment_ can be removed.

SergejMuhic:
> Then also the function _IfcCorrectObjectAssignment_ can be removed.

This was kind of redundant, wasn't it?

aothms:
> Then also the function IfcCorrectObjectAssignment can be removed.

The function then *should* be removed. It's no longer valid now with the type deleted. As well as the WR1 on IfcRelAssigns.

> so the sementically more correct type would be IfcStrippedOptional (invented for exactly this purpose - to mark a removed attribute while maintaining the attribute order.

Indeed. In the PR the attribute is deleted. This is not an option. So then indeed the most straight forward solution I see is to turn the attribute into an OPTIONAL IfcStrippedOptional.

We should find the time then to figure out what IfcStrippedOptional as a first class spec citizen looks like. I can see some advantages of e.g an empty entity paired with a rule that SIZEOF(IfcStrippedOptional) == 0 in the model.

anborr (approved):

anborr:
> > Then also the function IfcCorrectObjectAssignment can be removed.
> 
> The function then _should_ be removed. It's no longer valid now with the type deleted. As well as the WR1 on IfcRelAssigns.
> 
> > so the sementically more correct type would be IfcStrippedOptional (invented for exactly this purpose - to mark a removed attribute while maintaining the attribute order.
> 
> Indeed. In the PR the attribute is deleted. This is not an option. So then indeed the most straight forward solution I see is to turn the attribute into an OPTIONAL IfcStrippedOptional.
> 
> We should find the time then to figure out what IfcStrippedOptional as a first class spec citizen looks like. I can see some advantages of e.g an empty entity paired with a rule that SIZEOF(IfcStrippedOptional) == 0 in the model.

Then we would have to add IfcStrippedOptional to the Schema.

aothms:
Decision keep TYPE = IfcBoolean, but autogenenerate local rule `NOT EXISTS (attr name)` on the entity that references stripped optional.

larswik (approved):

pjanck (approved):
lftm

PR #467, opened 2022-07-29
Feature/property normalization

SergejMuhic:
I guess the name of this issue is not perfect 😄 but trying to keep the reference.

1. Rename IfcWorkTime.Start/Finish to StartDate/FinishDate
2. Change type IfcSweptDiskSolidPolygonal.FilletRadius to IfcNonNegativeLengthMeausure -> I cannot say I understand this change, especially with the description in place "The fillet that is equally applied to all transitions between the segments of the [IfcPolyline](http://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcPolyline.htm), providing the geometric representation for the Directrix. **If omitted, no fillet is applied to the segments**." What does FilletRadius = 0 compared to omitted mean?
3. Rename IfcStructuralCurveConnection.Axis to AxisDirection
4. Rename IfcClassification.Location to Specification
5. IfcIndexedPolyCurve.SelfIntersect mandatory and type to IfcLogical -> was this necessary? Again, especially with the description in place "Indication of whether the curve intersects itself or not; **this is for information only**."
6. Rename IfcMaterialRelationship.Expression to MaterialExpression
7. Rename IfcCurveStyleFontAndScaling.CurveFont to CurveStyleFont
8. Update of rule Name to reflect IfcClassification.Location name change

### DIFF
--- a/IFC4x3/Sections/Domain specific data schemas/Schemas/IfcStructuralAnalysisDomain/Entities/IfcStructuralCurveConnection/DocEntity.xml
+++ b/IFC4x3/Sections/Domain specific data schemas/Schemas/IfcStructuralAnalysisDomain/Entities/IfcStructuralCurveConnection/DocEntity.xml
@@ -7,7 +7,7 @@
 		<DocLocalization Locale="en" Name="Structural Curve Connection" />
 	</Localization>
 	<Attributes>
-		<DocAttribute Name="Axis" UniqueId="7ab9edfa-dde7-4c6c-8c6a-ce1cd8fc527c" DefinedType="IfcDirection" XsdFormat="attribute">
+		<DocAttribute Name="AxisDirection" UniqueId="7ab9edfa-dde7-4c6c-8c6a-ce1cd8fc527c" DefinedType="IfcDirection" XsdFormat="attribute">
 			<Documentation>Direction which is used in the definition of the local z axis.  _Axis_ is specified relative to the so-called global coordinate system, i.e. the _SELF\IfcProduct.ObjectPlacement_.
 
 &gt; NOTE&amp;nbsp; It is desirable and usually possible that many instances of _IfcStructuralCurveConnection_ and _IfcStructuralCurveMember_ share a common instance of _IfcDirection_ as their _Axis_ attribute.</Documentation>
--- a/IFC4x3/Sections/Resource definition data schemas/Schemas/IfcGeometricModelResource/Entities/IfcSweptDiskSolidPolygonal/DocEntity.xml
+++ b/IFC4x3/Sections/Resource definition data schemas/Schemas/IfcGeometricModelResource/Entities/IfcSweptDiskSolidPolygonal/DocEntity.xml
@@ -5,8 +5,11 @@
 		<DocLocalization Locale="en" Name="Swept Disk Solid Polygonal" />
 	</Localization>
 	<Attributes>
-		<DocAttribute Name="FilletRadius" UniqueId="1fa1bae9-ec32-470f-a6ca-750f269a3470" DefinedType="IfcPositiveLengthMeasure" AttributeFlags="1">
+		<DocAttribute Name="FilletRadius" UniqueId="1fa1bae9-ec32-470f-a6ca-750f269a3470" DefinedType="IfcNonNegativeLengthMeasure" AttributeFlags="1">
 			<Documentation>The fillet that is equally applied to all transitions between the segments of the _IfcPolyline_, providing the geometric representation for _the Directrix_. If omited, no fillet is applied to the segments.</Documentation>
+			<Definition>
+				<DocDefinitionRef xsi:nil="true" href="IfcNonNegativeLengthMeasure_0INRELfHT6d8CO87XOkxl" />
+			</Definition>
 		</DocAttribute>
 	</Attributes>
 	<WhereRules>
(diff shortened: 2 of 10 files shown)


TLiebich:
to 2.) agree, in this context being optional - meaning no radius (=0) and allowing radius=0 (i.e. having to options for the same) is not a semantic improvement - @aothms - revert that change?

to 5) we discussed it also in the ISO group and defended the change - see http://ifc43-docs.standards.buildingsmart.org/search?query=selfintersect - all other SelfIntersect attributes are mandatory logicals

SergejMuhic:
> to 2.) agree, in this context being optional - meaning no radius (=0) and allowing radius=0 (i.e. having to options for the same) is not a semantic improvement - @aothms - revert that change?
> 
> to 5) we discussed it also in the ISO group and defended the change - see http://ifc43-docs.standards.buildingsmart.org/search?query=selfintersect - all other SelfIntersect attributes are mandatory logicals

On 5. it would make sense to adjust the documentation then. Might I propose "Indication of whether the curve intersects itself or not; this is for information only. `UNKNOWN` shall be used also in exchanges where this information is deemed irrelevant."

SergejMuhic:
TODO: `IfcIndexedPolyline` and `IfcSweptDiskSolidPolygonal` was reverted. Implement here.

aothms:
2.

It's a unification of all FilletRadius types, e.g IfcIShapeProfileDef.FilletRadius which is also OPTIONAL IfcNonNegativeLengthMeasure.

The description there is

> The fillet between the web and the flange. 0 if sharp-edged, omitted if unknown.

I would propose to keep the unification and reword the documentation to:

> The fillet that is equally applied to all transitions between the segments of the Directrix. 0 if sharp-edged, omitted if unknown.

In that I also removed IfcPolyline because it can also be an IfcIndexedPolyCurve

SergejMuhic:
If the backwards compatibility rules are okay with this change, then I see no problems. The documentation change also works fine apart from removing `IfcPolyline` which is quite necessary 😄.

SergejMuhic:
@aothms @TLiebich It seems to me that the `IfcSweptDiskSolidPolygonal` was just changed to `IfcNonNegativeLengthMeausure` and the documentation has not changed. Must have been submitted to ISO like that then.



> It's a unification

On this I would have expected the unification to go the other way around honestly. From a more common entity to the one that is less used.

Are there plans on changing the documentation? I was waiting with this PR for that but now I would have to merge for Tunnel Deployment to have a solid baseline.

EDIT: Just remembered that the where rule and 0 filet radius do not fit well together either.

IFC4x3/Sections/Domain specific data schemas/Schemas/IfcStructuralAnalysisDomain/Entities/IfcStructuralCurveConnection/DocEntity.xml
```diff
@@ -7,7 +7,7 @@
 		<DocLocalization Locale="en" Name="Structural Curve Connection" />
 	</Localization>
 	<Attributes>
-		<DocAttribute Name="Axis" UniqueId="7ab9edfa-dde7-4c6c-8c6a-ce1cd8fc527c" DefinedType="IfcDirection" XsdFormat="attribute">
+		<DocAttribute Name="AxisDirection" UniqueId="7ab9edfa-dde7-4c6c-8c6a-ce1cd8fc527c" DefinedType="IfcDirection" XsdFormat="attribute">
 			<Documentation>Direction which is used in the definition of the local z axis.  _Axis_ is specified relative to the so-called global coordinate system, i.e. the _SELF\IfcProduct.ObjectPlacement_.
```
pjanck, line 11:
```suggestion
		<DocAttribute Name="AxisDirection" UniqueId="7ab9edfa-dde7-4c6c-8c6a-ce1cd8fc527c" DefinedType="IfcDirection" XsdFormat="attribute">
			<Documentation>Direction which is used in the definition of the local z axis.  _AxisDirection_ is specified relative to the so-called global coordinate system, i.e. the _SELF\IfcProduct.ObjectPlacement_.
```

Let's correct the documentation as well, if the attribute name got changed.

IFC4x3/Sections/Domain specific data schemas/Schemas/IfcStructuralAnalysisDomain/Entities/IfcStructuralCurveConnection/DocEntity.xml
```diff
@@ -7,7 +7,7 @@
...
 
 &gt; NOTE&amp;nbsp; It is desirable and usually possible that many instances of _IfcStructuralCurveConnection_ and _IfcStructuralCurveMember_ share a common instance of _IfcDirection_ as their _Axis_ attribute.</Documentation>
```
pjanck, line 13:
```suggestion
&gt; NOTE&amp;nbsp; It is desirable and usually possible that many instances of _IfcStructuralCurveConnection_ and _IfcStructuralCurveMember_ share a common instance of _IfcDirection_ as their _AxisDirection_ attribute.</Documentation>
```

Simile.

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcGeometricModelResource/Entities/IfcSweptDiskSolidPolygonal/DocEntity.xml
```diff
@@ -5,8 +5,11 @@
 		<DocLocalization Locale="en" Name="Swept Disk Solid Polygonal" />
 	</Localization>
 	<Attributes>
-		<DocAttribute Name="FilletRadius" UniqueId="1fa1bae9-ec32-470f-a6ca-750f269a3470" DefinedType="IfcPositiveLengthMeasure" AttributeFlags="1">
+		<DocAttribute Name="FilletRadius" UniqueId="1fa1bae9-ec32-470f-a6ca-750f269a3470" DefinedType="IfcNonNegativeLengthMeasure" AttributeFlags="1">
 			<Documentation>The fillet that is equally applied to all transitions between the segments of the _IfcPolyline_, providing the geometric representation for _the Directrix_. If omited, no fillet is applied to the segments.</Documentation>
```
pjanck, line 9:
```suggestion
			<Documentation>The fillet that is equally applied to all transitions between the segments of the _IfcPolyline_, providing the geometric representation for the _Directrix_. If omitted, no fillet is applied to the segments.</Documentation>
```

pjanck (changes requested):

larswik:
Suggest to merge and derive an issue both here and in the official repository regarding the fillet radius.

larswik (approved):

PR #468, opened 2022-07-29
Adding inverse attribute IfcGroup.ReferencedInStructures

SergejMuhic:
1. as title says


TLiebich:
see http://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcGroup.htm
only the description is missing

SergejMuhic:
> see http://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/lexical/IfcGroup.htm only the description is missing

Yep, waiting for the description, will add it when it is there.

anborr (approved):

aothms:
There's quite a few inverses without descriptions. E.g on product we have the same and also PositionedRelativeTo. On IfcRelReferencedInSpatialStructure actually no forward attribute has a description.

I'd say the forward attributes should definitely all be described. But for the inverses (subtypes of IfcRelationship at least) we can autogenerate sth like:

> Reference to the {entity} relationship, by which the {relationship entity other attribute description} can be obtained.

For IfcProduct.PositionedRelativeTo that would give:

> Reference to the IfcRelPositions relationship, by which the positioning element defining the source of the relative position can be obtained.

pjanck (approved):
lftm

PR #469, opened 2022-07-29
Feature/closed attribute on supertype

SergejMuhic:
1. Removing IfcPolygonalFaceSet.Closed
2. Removing IfcTriangulatedFaceSet.Closed
3. Adding IfcTessellatedFaceSet.Closed

I would like to add that this change is part of the issue:
-	Faces : LIST [1:?] OF IfcIndexedPolygonalFace;
+	Faces : SET [1:?] OF IfcIndexedPolygonalFace;

but it is not in the official documentation, so I am unsure on what to do with it.

### DIFF
--- a/IFC4x3/Sections/Resource definition data schemas/Schemas/IfcGeometricModelResource/DocSchema.xml
+++ b/IFC4x3/Sections/Resource definition data schemas/Schemas/IfcGeometricModelResource/DocSchema.xml
@@ -34,7 +34,7 @@
 				<DocDefinitionRef Name="IfcPlaneAngleMeasure" UniqueId="bdcf7f14-63da-4066-aa16-d68a69f41bc5" />
 				<DocDefinitionRef Name="IfcParameterValue" UniqueId="a3512312-a856-49a0-81ae-f9133eebaf2c" />
 				<DocDefinitionRef Name="IfcLengthMeasure" UniqueId="3c831c12-d959-431b-acb3-ead5264dd715" />
-				<DocDefinitionRef Name="IfcBoolean" UniqueId="5846255f-c88f-4014-be92-a94d7a3d3c39" />
+				<DocDefinitionRef id="IfcBoolean_1OHYLVo8z05BwIgKrwFJmv" Name="IfcBoolean" UniqueId="5846255f-c88f-4014-be92-a94d7a3d3c39" />
 				<DocDefinitionRef Name="IfcInteger" UniqueId="10de1fa3-b124-4d41-84fa-113459d99491" />
 				<DocDefinitionRef Name="IfcPositiveInteger" UniqueId="6b5c4b68-240e-438f-a05a-e1305da435ec" />
 				<DocDefinitionRef Name="IfcLabel" UniqueId="07fd9c74-8190-4dc9-bee0-d244ed13b0f8" />
--- a/IFC4x3/Sections/Resource definition data schemas/Schemas/IfcGeometricModelResource/Entities/IfcTessellatedFaceSet/DocEntity.xml
+++ b/IFC4x3/Sections/Resource definition data schemas/Schemas/IfcGeometricModelResource/Entities/IfcTessellatedFaceSet/DocEntity.xml
@@ -14,6 +14,12 @@
 		<DocAttribute Name="Coordinates" UniqueId="28b00035-2a37-4bda-aa0e-7845352fa060" DefinedType="IfcCartesianPointList3D" XsdFormat="attribute">
 			<Documentation>An ordered list of Cartesian points used by the coordinate index defined at the subtypes of _IfcTessellatedFaceSet_.</Documentation>
 		</DocAttribute>
+		<DocAttribute Name="Closed" UniqueId="0ed95f80-d2a4-4366-9e14-879da9de6926" DefinedType="IfcBoolean">
+			<Documentation>Indication whether the _IfcPolygonalFaceSet_ is a closed shell or not. If omited no such information can be provided.</Documentation>
+			<Definition>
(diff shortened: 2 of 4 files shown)


TLiebich:
regarding:

    Removing IfcPolygonalFaceSet.Closed
    Removing IfcTriangulatedFaceSet.Closed
    Adding IfcTessellatedFaceSet.Closed

we had to revert this change in order to keep attribute order, since Closed has a different position in the two subtypes.

regarding SET vs. LIST - as the name suggest, it should have been a SET, but it is a LIST in IFC4. Not sure, whether we should make this change in IFC4.3 at the late point in time. despite being very minror.

SergejMuhic:
> regarding:
> 
> ```
> Removing IfcPolygonalFaceSet.Closed
> Removing IfcTriangulatedFaceSet.Closed
> Adding IfcTessellatedFaceSet.Closed
> ```
This seemed like a big change indeed. I guess this makes the PR irrelevant then.
> 
> we had to revert this change in order to keep attribute order, since Closed has a different position in the two subtypes.
> 
> regarding SET vs. LIST - as the name suggest, it should have been a SET, but it is a LIST in IFC4. Not sure, whether we should make this change in IFC4.3 at the late point in time. despite being very minror.

You probably mean the name of the entity, right? I was looking for "set" in the attribute. 😄

Set definitely makes more sense but it would also work with just adding the UNIQUE keyword to the list, thereby keeping the order of elements. Can't think of a scenario where order is required but looking at the original definition, it might be that other people have seen benefits of an ordered collection of faces.

Also, looking at the referenced **indexedFaceSet** from ISO 19775-1, it did have ordered coords (not faces) due to not having a separate face (again as stated above faces might not benefit from order) definition, although the documentation and name imply set. Indexed is also part of the name so it is a bit of a conundrum what is more important (set being unordered). I have not looked into the documentation deeply. Order might be important though.

Maybe we are putting too much emphasis on naming?

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcGeometricModelResource/DocSchema.xml
```diff
@@ -34,7 +34,7 @@
 				<DocDefinitionRef Name="IfcPlaneAngleMeasure" UniqueId="bdcf7f14-63da-4066-aa16-d68a69f41bc5" />
 				<DocDefinitionRef Name="IfcParameterValue" UniqueId="a3512312-a856-49a0-81ae-f9133eebaf2c" />
 				<DocDefinitionRef Name="IfcLengthMeasure" UniqueId="3c831c12-d959-431b-acb3-ead5264dd715" />
-				<DocDefinitionRef Name="IfcBoolean" UniqueId="5846255f-c88f-4014-be92-a94d7a3d3c39" />
+				<DocDefinitionRef id="IfcBoolean_1OHYLVo8z05BwIgKrwFJmv" Name="IfcBoolean" UniqueId="5846255f-c88f-4014-be92-a94d7a3d3c39" />
```
anborr, line 37: I don't understand this change. Is it really necessary?

anborr (changes requested):
The documentation must be changed as well to be consistent.
Currently, IfcPolygonalFaceSet and IfcTriangulatedFaceSet have the following doc which must be lifted to IfcTessellatedFaceSet:

Informal Propositions:

If Closed = TRUE, the same topological constraints as imposed by an faceted boundary representation shall be safisfied. In particular:
-         each face shall be unique;
-         each edge shall be used by exactly two boundaries of faces;
-         the topological normal at each point on the boundary of the triangulated face set shall point away from the material of the solid.

SergejMuhic:
We keep LIST and add UNIQUE.

pjanck:
What is the state of:

> I guess this makes the PR irrelevant then.

> We keep LIST and add UNIQUE.

> Lifting the Informal Propositions to the supertype as mentioned by @anborr.

IFC4x3/Sections/Resource definition data schemas/Schemas/IfcGeometricModelResource/Entities/IfcTessellatedFaceSet/DocEntity.xml
```diff
@@ -14,6 +14,12 @@
 		<DocAttribute Name="Coordinates" UniqueId="28b00035-2a37-4bda-aa0e-7845352fa060" DefinedType="IfcCartesianPointList3D" XsdFormat="attribute">
 			<Documentation>An ordered list of Cartesian points used by the coordinate index defined at the subtypes of _IfcTessellatedFaceSet_.</Documentation>
 		</DocAttribute>
+		<DocAttribute Name="Closed" UniqueId="0ed95f80-d2a4-4366-9e14-879da9de6926" DefinedType="IfcBoolean">
+			<Documentation>Indication whether the _IfcPolygonalFaceSet_ is a closed shell or not. If omited no such information can be provided.</Documentation>
```
pjanck:
```suggestion
			<Documentation>Indication whether the _IfcTessellatedFaceSet_ is a closed shell or not. If omitted no such information can be provided.</Documentation>
```

pjanck (changes requested):

SergejMuhic:
> What is the state of:
> 
> > I guess this makes the PR irrelevant then.
> 
> > We keep LIST and add UNIQUE.
> 
> > Lifting the Informal Propositions to the supertype as mentioned by @anborr.

I guess I can just close this PR as I did not change the LIST to UNIQUE and make a new PR. I will make a new PR for that.

Regarding lifting informal propositions, I would oppose as the attribute `Closed` is on the subtypes once again when reverting.

PR #493, opened 2023-01-23
jwg12 IfcGetBasisSurface

SergejMuhic:
1. updated according to bSI instructions

EDIT: I would have at least preferred an "else if". Then you would have an easier time deciding whether to interpret it as a "switch case" or "else if".

### DIFF
--- a/IFC4x3/Sections/Resource definition data schemas/Schemas/IfcGeometryResource/Functions/IfcGetBasisSurface/Expression.txt
+++ b/IFC4x3/Sections/Resource definition data schemas/Schemas/IfcGeometryResource/Functions/IfcGetBasisSurface/Expression.txt
@@ -21,10 +21,22 @@
     (* For an IfcCompositeCurveOnSurface the BasisSurface is the intersection of the BasisSurface of all the segments. *)
 
     N := SIZEOF(C\IfcCompositeCurve.Segments);
-    Surfs := IfcGetBasisSurface(C\IfcCompositeCurve.Segments[1].ParentCurve);
+
+    IF ('IFCGEOMETRYRESOURCE.IFCCURVESEGMENT' IN TYPEOF(C\IfcCompositeCurve.Segments[1])) THEN
+      Surfs := IfcGetBasisSurface(C\IfcCompositeCurve.Segments[1]\IfcCurveSegment.ParentCurve);
+    END_IF;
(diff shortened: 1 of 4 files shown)


IFC4x3/Sections/Resource definition data schemas/Schemas/IfcGeometryResource/Functions/IfcGetBasisSurface/Expression.txt
```diff
@@ -21,10 +21,22 @@
...
+
     IF N > 1 THEN
       REPEAT i := 2 TO N;
-        Surfs := Surfs * IfcGetBasisSurface(C\IfcCompositeCurve.Segments[1].ParentCurve);
+        IF ('IFCGEOMETRYRESOURCE.IFCCURVESEGMENT' IN TYPEOF(C\IfcCompositeCurve.Segments[i])) THEN
+          Surfs := Surfs * IfcGetBasisSurface(C\IfcCompositeCurve.Segments[i]\IfcCurveSegment.ParentCurve);
+        END_IF;
+        IF ('IFCGEOMETRYRESOURCE.IFCCOMPOSITECURVESEGMENT' IN TYPEOF(C\IfcCompositeCurve.Segments[i])) THEN
+          Surfs := Surfs * IfcGetBasisSurface(C\IfcCompositeCurve.Segments[i]\IfcCompositeCurveSegment.ParentCurve);
+        END_IF;
```
pjanck, line 39: In deleted line, there is `Segments[1]`, in the additions, it is `Segments[i]`. Intentional fix?

SergejMuhic: Yes, that was a mistake to be corrected if I remember correctly. Old one.

pjanck (approved):
Not the nicest of syntax (agreeing with the edit note), but so be it.

SergejMuhic:
@TLiebich any opinions? I think we should at least make it an "If else if". But if we do it here, we have to do it on bSI also. I am in no position to do that.

TLiebich:
I would leave it as is (of cause including the differentiation of curve segement and composite curve segment. We inherited it from STEP, I just checked the late(st) publication and it is not altered there.

PR #527, opened 2023-03-01
 Remove Name from IfcVoxelData

### LINKED ISSUE #499: IfcVoxelData Name Attribute
https://github.com/bSI-InfraRoom/IFC-Specification/blob/95f944646991624619ad3f80fb53793d650319dc/IFC4x3/Sections/Core%20data%20schemas/Schemas/IfcProductExtension/Entities/IfcVoxelData/DocEntity.xml

IfcVoxelData inherits from IfcRoot, but has an explicit attribute Name in both classes.  It's not derived.
This seems wrong and causes a defect in the published documentation (ie attribute 3 is missing in inherited attributes).

https://bsi-infraroom.github.io/IFC-Documentation-Tunnel/4_4_0_0/general/HTML/

### LINKED ISSUE #525: `IfcVoxelData` has Name attribute
This somehow slipped through. Needs to be removed. I am getting issues when intepreting.

SergejMuhic:
Fixes https://github.com/bSI-InfraRoom/IFC-Specification/issues/525
Fixes https://github.com/bSI-InfraRoom/IFC-Specification/issues/499

1. Removed Name attribute


pjanck (approved):
lftm

larswik (approved):

PR #532, opened 2023-03-05
`IfcVoxelData` Value type and WRs

### LINKED ISSUE #492: IfcVoxelData indeterminate array bounds not allowed
Purely technical issue outside of the bigger discussion on voxels in general:

```
ENTITY IfcIntegerVoxelData
 SUBTYPE OF (IfcVoxelData);
	Values : ARRAY [1:?] OF IfcInteger;
	Unit : OPTIONAL IfcUnit;
END_ENTITY;
```

> *8.2 Aggregation data types*
>
> An `array` is a fixed-size ordered collection. It is indexed by a sequence of integers.

or if we fast forward to 8.2.1 more explicitly:

> *8.2.1 Array data type*
> 
> ...
>
> NOTE 1 The bounds may be positive, negative or zero, but may not be indeterminate (`?`) (see 14.2).

SergejMuhic:
Fixes #492 

1. [eof lines added](https://github.com/bSI-InfraRoom/IFC-Specification/commit/ce087396bc5a3949e2b90f823dc9464048ad2794)
2. [adding derived GridSize](https://github.com/bSI-InfraRoom/IFC-Specification/commit/26f44afdae8354079eca7620c67ad10e301f4ad5)
3. [upper bound of Values attribute](https://github.com/bSI-InfraRoom/IFC-Specification/commit/4ac73b71b9bf1914e3bf2dfc24b2fc910831522b)
4. [adding HasRepresentation WR](https://github.com/bSI-InfraRoom/IFC-Specification/commit/5ab993743d0e42110bbfdc757af7ce223a5dcd3e)
5. [adding VoxelGridRepresentation WR](https://github.com/bSI-InfraRoom/IFC-Specification/commit/e3ebd444ba8f07d25d2989bc3a573674999796d5)
6. [adding SameGeometricRepresentation WR](https://github.com/bSI-InfraRoom/IFC-Specification/commit/35c0e02285498d10073ea98efa97a6c5678e083e)

I realized that the WRs should be on `IfcVoxelData` probably. Not that important and quickly fixed so please ignore this.

EDIT: Forgot to add before. The issue with GridSize is that it is not known at object construction time. Any suggestions how to handle this?

### DIFF
--- a/IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
+++ b/IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
@@ -14,21 +14,5 @@
 			</Definition>
 		</DocAttribute>
 	</Attributes>
-	<WhereRules>
-		<DocWhereRule Name="HasRepresentation" UniqueId="16865dc1-1980-43fd-8290-4bac0e77c2ca">
-			<Documentation>_IfcIntegerVoxelData_._Representation_ shall exist.</Documentation>
-			<Expression>EXISTS(SELF\IfcProduct.Representation)</Expression>
-		</DocWhereRule>
-		<DocWhereRule Name="VoxelGridRepresentation" UniqueId="bb6f7739-0fb3-4899-a086-823cc53156f2">
-			<Documentation>The geometric representation item in _IfcProductDefinitionShape_._Representations_ shall be _IfcVoxelGrid_.</Documentation>
-			<Expression>&apos;IFCVOXELGRID&apos; IN TYPEOF(SELF\IfcProduct.Representation.Representations[1].Items[1])</Expression>
-		</DocWhereRule>
-		<DocWhereRule Name="SameGeometricRepresentation" UniqueId="9f6abff4-3bdb-4be7-8106-237e3b2e0e47">
-			<Documentation>RelatingProduct shall have the same geometric representation.</Documentation>
-			<Expression>EXISTS(SELF\IfcObjectDefinition.HasAssignments\IfcRelAssignsToProduct.RelatingProduct.Representation)
-AND
-SELF\IfcProduct.Representation.Representations[1].Items[1] = SELF\IfcObjectDefinition.HasAssignments\IfcRelAssignsToProduct.RelatingProduct.Representation.Representations[1].Items[1]</Expression>
-		</DocWhereRule>
-	</WhereRules>
 </DocEntity>
 
--- a/IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
+++ b/IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <DocEntity xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" Name="IfcIntegerVoxelData" UniqueId="f3d953cb-24b7-483d-85e1-d44c43d4d22b" BaseDefinition="IfcVoxelData" EntityFlags="32">
 	<Attributes>
-		<DocAttribute Name="Values" UniqueId="342aa966-eaa4-4d0c-87e5-91988a4f1c0a" DefinedType="IfcInteger" AggregationType="2" AggregationLower="1">
+		<DocAttribute Name="Values" UniqueId="342aa966-eaa4-4d0c-87e5-91988a4f1c0a" DefinedType="IfcInteger" AggregationType="2" AggregationLower="1" AggregationUpper="GridSize">
 			<Documentation>The values assigned to the voxels. First x, then y and lastly z.</Documentation>
 		</DocAttribute>
 		<DocAttribute Name="Unit" UniqueId="d23d69e2-c35a-41d5-bf3b-9533670b9eb7" DefinedType="IfcUnit" AttributeFlags="1">
--- a/IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
+++ b/IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
@@ -25,9 +25,9 @@
 		</DocWhereRule>
 		<DocWhereRule Name="SameGeometricRepresentation" UniqueId="9f6abff4-3bdb-4be7-8106-237e3b2e0e47">
 			<Documentation>RelatingProduct shall have the same geometric representation.</Documentation>
-			<Expression>EXISTS(SELF\IfcObjectDefinition.HasAssignment\IfcRelAssignsToProduct.RelatingProduct.Representation)
+			<Expression>EXISTS(SELF\IfcObjectDefinition.HasAssignments\IfcRelAssignsToProduct.RelatingProduct.Representation)
 AND
-SELF\IfcProduct.Representation.Representations[1].Items[1] = SELF\IfcObjectDefinition.HasAssignment\IfcRelAssignsToProduct.RelatingProduct.Representation.Representations[1].Items[1]</Expression>
+SELF\IfcProduct.Representation.Representations[1].Items[1] = SELF\IfcObjectDefinition.HasAssignments\IfcRelAssignsToProduct.RelatingProduct.Representation.Representations[1].Items[1]</Expression>
 		</DocWhereRule>
 	</WhereRules>
 </DocEntity>
--- a/IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
+++ b/IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
@@ -7,6 +7,12 @@
 		<DocAttribute Name="Unit" UniqueId="d23d69e2-c35a-41d5-bf3b-9533670b9eb7" DefinedType="IfcUnit" AttributeFlags="1">
 			<Documentation>An optional unit for the integer values.</Documentation>
 		</DocAttribute>
+		<DocAttribute Name="GridSize" UniqueId="6f8d6a89-e211-4ca1-a678-e4221b561f42" DefinedType="IfcInteger">
+			<Derived> SIZEOF(SELF\IfcProduct.Representation.Representations[1].Items[1]\IfcVoxelGrid.Voxels)</Derived>
+			<Definition>
+				<DocDefinitionRef xsi:nil="true" href="IfcInteger_3wu0Hj__nD9vk_ZJIPfGUc" />
+			</Definition>
+		</DocAttribute>
 	</Attributes>
 </DocEntity>
 
--- a/IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
+++ b/IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
@@ -19,6 +19,10 @@
 			<Documentation>_IfcIntegerVoxelData_. _Representation_ shall exist.</Documentation>
 			<Expression>EXISTS(SELF\IfcProduct.Representation)</Expression>
 		</DocWhereRule>
+		<DocWhereRule Name="VoxelGridRepresentation" UniqueId="bb6f7739-0fb3-4899-a086-823cc53156f2">
+			<Documentation>The geometric representation item in _IfcProductDefinitionShape_._Representations_ shall be _IfcVoxelGrid_.</Documentation>
+			<Expression>&apos;IFCVOXELGRID&apos; IN TYPEOF(SELF\IfcProduct.Representation.Representations[1].Items[1])</Expression>
+		</DocWhereRule>
 	</WhereRules>
 </DocEntity>
 
--- a/IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
+++ b/IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
@@ -16,7 +16,7 @@
 	</Attributes>
 	<WhereRules>
 		<DocWhereRule Name="HasRepresentation" UniqueId="16865dc1-1980-43fd-8290-4bac0e77c2ca">
-			<Documentation>_IfcIntegerVoxelData_. _Representation_ shall exist.</Documentation>
+			<Documentation>_IfcIntegerVoxelData_._Representation_ shall exist.</Documentation>
 			<Expression>EXISTS(SELF\IfcProduct.Representation)</Expression>
 		</DocWhereRule>
 		<DocWhereRule Name="VoxelGridRepresentation" UniqueId="bb6f7739-0fb3-4899-a086-823cc53156f2">
--- a/IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
+++ b/IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
@@ -14,5 +14,11 @@
 			</Definition>
 		</DocAttribute>
 	</Attributes>
+	<WhereRules>
+		<DocWhereRule Name="HasRepresentation" UniqueId="16865dc1-1980-43fd-8290-4bac0e77c2ca">
+			<Documentation>_IfcIntegerVoxelData_. _Representation_ shall exist.</Documentation>
+			<Expression>EXISTS(SELF\IfcProduct.Representation)</Expression>
+		</DocWhereRule>
+	</WhereRules>
 </DocEntity>
 
--- a/IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
+++ b/IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
@@ -23,6 +23,12 @@
 			<Documentation>The geometric representation item in _IfcProductDefinitionShape_._Representations_ shall be _IfcVoxelGrid_.</Documentation>
 			<Expression>&apos;IFCVOXELGRID&apos; IN TYPEOF(SELF\IfcProduct.Representation.Representations[1].Items[1])</Expression>
 		</DocWhereRule>
+		<DocWhereRule Name="SameGeometricRepresentation" UniqueId="9f6abff4-3bdb-4be7-8106-237e3b2e0e47">
+			<Documentation>RelatingProduct shall have the same geometric representation.</Documentation>
+			<Expression>EXISTS(SELF\IfcObjectDefinition.HasAssignment\IfcRelAssignsToProduct.RelatingProduct.Representation)
+AND
+SELF\IfcProduct.Representation.Representations[1].Items[1] = SELF\IfcObjectDefinition.HasAssignment\IfcRelAssignsToProduct.RelatingProduct.Representation.Representations[1].Items[1]</Expression>
+		</DocWhereRule>
 	</WhereRules>
 </DocEntity>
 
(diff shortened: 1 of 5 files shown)


IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
```diff
@@ -1,12 +1,34 @@
...
+		<DocAttribute Name="GridSize" UniqueId="6f8d6a89-e211-4ca1-a678-e4221b561f42" DefinedType="IfcInteger">
+			<Derived> SIZEOF(SELF\IfcProduct.Representation.Representations[1].Items[1]\IfcVoxelGrid.Voxels)</Derived>
+			<Definition>
+				<DocDefinitionRef xsi:nil="true" href="IfcInteger_3wu0Hj__nD9vk_ZJIPfGUc" />
+			</Definition>
+		</DocAttribute>
 	</Attributes>
+	<WhereRules>
+		<DocWhereRule Name="HasRepresentation" UniqueId="16865dc1-1980-43fd-8290-4bac0e77c2ca">
+			<Documentation>_IfcIntegerVoxelData_. _Representation_ shall exist.</Documentation>
```
pjanck:
```suggestion
			<Documentation>_IfcIntegerVoxelData_._Representation_ shall exist.</Documentation>
```

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
```diff
@@ -1,12 +1,34 @@
...
 	<Attributes>
-		<DocAttribute Name="Values" UniqueId="342aa966-eaa4-4d0c-87e5-91988a4f1c0a" DefinedType="IfcInteger" AggregationType="2" AggregationLower="1">
+		<DocAttribute Name="Values" UniqueId="342aa966-eaa4-4d0c-87e5-91988a4f1c0a" DefinedType="IfcInteger" AggregationType="2" AggregationLower="1" AggregationUpper="GridSize">
 			<Documentation>The values assigned to the voxels. First x, then y and lastly z.</Documentation>
 		</DocAttribute>
 		<DocAttribute Name="Unit" UniqueId="d23d69e2-c35a-41d5-bf3b-9533670b9eb7" DefinedType="IfcUnit" AttributeFlags="1">
 			<Documentation>An optional unit for the integer values.</Documentation>
 		</DocAttribute>
+		<DocAttribute Name="GridSize" UniqueId="6f8d6a89-e211-4ca1-a678-e4221b561f42" DefinedType="IfcInteger">
+			<Derived> SIZEOF(SELF\IfcProduct.Representation.Representations[1].Items[1]\IfcVoxelGrid.Voxels)</Derived>
```
pjanck, line 11: 1. There are a lot of `[1]` getters. The `GridSize` derive would fail with:

- two `Representations`, and
- `IfcVoxelGrid` being in the second, 

although the information is provided. 

2. `IfcRepresentation.Items` is a SET - accessing element at `[1]` doesn't work.

I believe, the expression should be refactored.

pjanck (changes requested):

pjanck:
> The issue with GridSize is that it is not known at object construction time. Any suggestions how to handle this?

Idea: Doubling of information (explicit GridSize attribute as IfcInteger, that is serialized) with a Where rule ensuring validity (that GridSize equals the formula currently in Derive expression)?

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
SergejMuhic: The dot is supposed to be an "operator" for the path to the `Representation` attribute. 😄 

EDIT: retract comment. I have missed the point.

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
SergejMuhic: Just continuing on the IFC patterns. To my knowledge, there are no other getters for `SET`.

Not sure I can agree with the case of two representations if `IfcVoxelGrid` is in the first (whatever that means in a `SET`). You were probably aiming at fetching from a `SET` with two member though, right? In that case I understand.

The only other option I can think of is a `REPEAT` to iterate over representations and find the one with `IfcVoxelGrid`.

Your comments are not taking into account the `VoxelGridRepresentation` WR. I could also restrict SIZEOF Items to 1 and the [1] item to `IfcVoxelGrid`. That would make it clearer but playing around with `GridSize` does not seem to be productive IMHO.

SergejMuhic:
> > The issue with GridSize is that it is not known at object construction time. Any suggestions how to handle this?
> 
> Idea: Doubling of information (explicit GridSize attribute as IfcInteger, that is serialized) with a Where rule ensuring validity (that GridSize equals the formula currently in Derive expression)?

Yup, this could be the way to go.

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
pjanck: > You were probably aiming at fetching from a SET with two member though, right? In that case I understand.

Correct.

> The only other option I can think of is a REPEAT to iterate over representations and find the one with IfcVoxelGrid.

This seems a reasonable solution.

> Your comments are not taking into account the VoxelGridRepresentation WR. I could also restrict SIZEOF Items to 1 and the [1] item to IfcVoxelGrid.

The current state of the expression is:

    'IFCVOXELGRID' IN TYPEOF(SELF\IfcProduct.Representation.Representations[1].Items[1])

Expanding it to ensure size to be 1 would solve the issue; however, I believe that limiting to a single representation would be very restrictive. Is the requirement "there should be one representation of type `IfcVoxelGrid`" or "there should be *only* one representation of type `IfcVoxelGrid`"?

larswik: > Expanding it to ensure size to be 1 would solve the issue; however, I believe that limiting to a single representation would be very restrictive. Is the requirement "there should be one representation of type `IfcVoxelGrid`" or "there should be _only_ one representation of type `IfcVoxelGrid`"?

To me it would feel logical that an IfcVoxelData instance should have exactly one representation and it should be of type IfcVoxelGrid. On the other hand, there could be several IfcVoxelData sharing the same IfcVoxelGrid.

pjanck:
> Yup, this could be the way to go.

Pinging @larswik and @SergejMuhic - let's get this moving again (especially following #617 and #618 .

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
pjanck: > an IfcVoxelData instance should have exactly one representation and it should be of type IfcVoxelGrid

I can live with that. We can include this in #618 as well (e.g. see [here](https://github.com/bSI-InfraRoom/IFC-Specification/blob/f8e8310e0ca8bc19b50ac40f4bf8b37146494c2b/IFC4x3/Sections/Resource%20definition%20data%20schemas/Schemas/IfcRepresentationResource/Functions/IfcShapeRepresentationTypes/Expression.txt#L191-L198)).

aothms: To me this derived attribute is more evidence that voxels as used in 4.4 shouldn't be burried deeply into the representation item, but should really be in the property domain. If these kind of direct dependencies are added it's clear that this is not a representation in line of how representations in IFC are currently used.

`set`s can be indexed numerically, but the only guarantee is that the indexing is stable. Index `n` remains index `n` for the lifetime of the `set`.

--

I also don't really understand the meaning of this attribute. In the IfcVoxelGrid there is `VoxelSizeX`,  `NumberOfVoxelsX` and `Voxels`. The last one is a 1-d flat sequence of booleans does that really make sense to expose upwards? It seems to me like the least meaningful attribute for an end-user?

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
```diff
@@ -1,12 +1,34 @@
...
+			<Documentation>_IfcIntegerVoxelData_._Representation_ shall exist.</Documentation>
+			<Expression>EXISTS(SELF\IfcProduct.Representation)</Expression>
+		</DocWhereRule>
+		<DocWhereRule Name="VoxelGridRepresentation" UniqueId="bb6f7739-0fb3-4899-a086-823cc53156f2">
+			<Documentation>The geometric representation item in _IfcProductDefinitionShape_._Representations_ shall be _IfcVoxelGrid_.</Documentation>
+			<Expression>&apos;IFCVOXELGRID&apos; IN TYPEOF(SELF\IfcProduct.Representation.Representations[1].Items[1])</Expression>
+		</DocWhereRule>
+		<DocWhereRule Name="SameGeometricRepresentation" UniqueId="9f6abff4-3bdb-4be7-8106-237e3b2e0e47">
+			<Documentation>RelatingProduct shall have the same geometric representation.</Documentation>
+			<Expression>EXISTS(SELF\IfcObjectDefinition.HasAssignment\IfcRelAssignsToProduct.RelatingProduct.Representation)
```
aothms: Isn't this `HasAssignments` (plural) and therefore should also be a loop of some sorts?

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
```diff
@@ -1,12 +1,34 @@
...
+AND
+SELF\IfcProduct.Representation.Representations[1].Items[1] = SELF\IfcObjectDefinition.HasAssignment\IfcRelAssignsToProduct.RelatingProduct.Representation.Representations[1].Items[1]</Expression>
```
aothms: I think this is way too restrictive. The implication of other forms of assignments (not RelAssignsToProduct) or other types of assignees (RelatingProduct is a select) could be better assessed (this should probably be a function with proper type tests). But in general I would say these kind of constraints that span many domains should not be embedded in the schema, because they really violate the modularity and composability of the schema.

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
SergejMuhic: > To me this derived attribute is more evidence that voxels as used in 4.4 shouldn't be burried deeply into the representation item, but should really be in the property domain.

Voxels are used as geometry items in software. We also have presentation requirements for them, so the definition that fits most is _IfcGeometricRepresentationItem_.

> I also don't really understand the meaning of this attribute. In the IfcVoxelGrid there is `VoxelSizeX`, `NumberOfVoxelsX` and `Voxels`.

It is in the same ball park as Dim. Meta information that is derived from somewhere else. I do agree that extracting it from geometry is unorthodox but I would like to see geometry used more extensively for generating information from the model.

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
SergejMuhic: Correct.

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
SergejMuhic: > (RelatingProduct is a select) could be better assessed (this should probably be a function with proper type tests).

Right.

> But in general I would say these kind of constraints that span many domains should not be embedded in the schema, because they really violate the modularity and composability of the schema.

I do not fully agree here. I agree to a certain degree. `IfcIndexedColourMap` also relies on the indices of `IfcTriangulatedFaceSet`, but it is not encoded as a rule. This is just a consequence of working with indexed data structures. I wouldn't limit this kind of notion for connecting data.

Do you think we should just drop the rule or even the complete definition of VoxelData?

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
aothms: One more thing. I'm not an expert at IfcDoc definitions, but does this actually result in an OPTIONAL  element specification?

```
Values : ARRAY [1:GridSize] OF OPTIONAL IfcInteger;
                               ^^^^^^^^
```

aothms: And yet one more thing. This only fixes IfcIntegerVoxelData, but there are more right?

SergejMuhic: Yes, before merging, we would need to address all other subtypes of `IfcVoxelData` also.

anborr (approved):

larswik:
It has been a while, and I'm trying to recapitulate.

**These are fixed from what I understand:**
- The ARRAY[1:GridSize]
- Ifc*VoxelData shall have exactly one representation which shall be of type IfcVoxelGrid

**Still discussed (see attached image)?:**
- The restriction on HasAssignments. i am not sure about the syntax, but is this safe?:  `SELF\IfcObjectDefinition.HasAssignments\IfcRelAssignsToProduct`
  - I assume that there must be an assigned IfcProduct (i.e. IfcRelAssignsToProduct) among the assignments for Ifc*VoxelData for a voxel grid dataset to make sense. Can there be other types of assignments as well?
- I also assume that the assigned IfcProduct can have several representation items, but one of them must be exactly the same representation item (The IfcVoxelGrid) as the Ifc*VoxelData representation?
- If the previous point is true, and/or the first point is unsafe then the where rule SameGeometricRepresentation should be updated accordingly.
image

Ping @SergejMuhic and @pjanck

SergejMuhic:
- it should be safe because if the value to the left of full stop is indeterminate, it evaluates to indeterminate, EXISTS then returns false.
- this is something to discuss. Not sure whether we want to allow this. If yes, then the rule has to be adjusted.
- right.

larswik:
> * it should be safe because if the value to the left of full stop is indeterminate, it evaluates to indeterminate, EXISTS then returns false.

Ok, fine!

> * this is something to discuss. Not sure whether we want to allow this. If yes, then the rule has to be adjusted.

Maybe discuss in the project meeting tonight? I would like to see domain requirements for this. Personally, I think that many representations (where IfcVoxelGrid is one), should be allowed schema wise (and maybe restricted in an MVD), as for any product. But I'm not a domain expert.

> * right.

larswik (approved):
Once IfcIntegerVoxelData is cleaned up this LFTM!

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
```diff
@@ -1,12 +1,34 @@
...
 <DocEntity xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" Name="IfcIntegerVoxelData" UniqueId="f3d953cb-24b7-483d-85e1-d44c43d4d22b" BaseDefinition="IfcVoxelData" EntityFlags="32">
 	<Attributes>
-		<DocAttribute Name="Values" UniqueId="342aa966-eaa4-4d0c-87e5-91988a4f1c0a" DefinedType="IfcInteger" AggregationType="2" AggregationLower="1">
+		<DocAttribute Name="Values" UniqueId="342aa966-eaa4-4d0c-87e5-91988a4f1c0a" DefinedType="IfcInteger" AggregationType="2" AggregationLower="1" AggregationUpper="GridSize">
 			<Documentation>The values assigned to the voxels. First x, then y and lastly z.</Documentation>
 		</DocAttribute>
 		<DocAttribute Name="Unit" UniqueId="d23d69e2-c35a-41d5-bf3b-9533670b9eb7" DefinedType="IfcUnit" AttributeFlags="1">
 			<Documentation>An optional unit for the integer values.</Documentation>
 		</DocAttribute>
+		<DocAttribute Name="GridSize" UniqueId="6f8d6a89-e211-4ca1-a678-e4221b561f42" DefinedType="IfcInteger">
```
pjanck, line 10:
```suggestion
		<DocAttribute Name="GridSize" UniqueId="6f8d6a89-e211-4ca1-a678-e4221b561f42" DefinedType="IfcPositiveInteger">
```

Propose to make it a *positive* integer.

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
```diff
@@ -1,12 +1,34 @@
...
+		</DocWhereRule>
+		<DocWhereRule Name="VoxelGridRepresentation" UniqueId="bb6f7739-0fb3-4899-a086-823cc53156f2">
+			<Documentation>The geometric representation item in _IfcProductDefinitionShape_._Representations_ shall be _IfcVoxelGrid_.</Documentation>
+			<Expression>&apos;IFCVOXELGRID&apos; IN TYPEOF(SELF\IfcProduct.Representation.Representations[1].Items[1])</Expression>
+		</DocWhereRule>
+		<DocWhereRule Name="SameGeometricRepresentation" UniqueId="9f6abff4-3bdb-4be7-8106-237e3b2e0e47">
+			<Documentation>RelatingProduct shall have the same geometric representation.</Documentation>
+			<Expression>EXISTS(SELF\IfcObjectDefinition.HasAssignments\IfcRelAssignsToProduct.RelatingProduct.Representation)
+AND
+SELF\IfcProduct.Representation.Representations[1].Items[1] = SELF\IfcObjectDefinition.HasAssignments\IfcRelAssignsToProduct.RelatingProduct.Representation.Representations[1].Items[1]</Expression>
```
pjanck: - `HasAssignments` is a collection, some sort of `[1]` accessor is missing.
- `IfcRelAssignsToProduct.RelatingProduct` is a select, a cast is missing.

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcVoxelData/DocEntity.xml
```diff
@@ -5,4 +5,30 @@
...
+AND
+&apos;IFCRELASSIGNSTOPRODUCT&apos; IN TYPEOF(SELF\IfcObjectDefinition.HasAssignments[1])</Expression>
+		</DocWhereRule>
+		<DocWhereRule Name="HasRepresentation" UniqueId="3caf1533-687f-4af7-9621-967ad14407a6">
+			<Documentation>_IfcIntegerVoxelData_._Representation_ shall exist.</Documentation>
+			<Expression>EXISTS(SELF\IfcProduct.Representation)</Expression>
+		</DocWhereRule>
+		<DocWhereRule Name="SameRepresentation" UniqueId="02d00e6c-4372-463f-aef6-49b483f51e21">
+			<Documentation>RelatingProduct shall have the same product representation.</Documentation>
+			<Expression>SELF\IfcProduct.Representation = SELF\IfcObjectDefinition.HasAssignments\IfcRelAssignsToProduct.RelatingProduct.Representation</Expression>
```
pjanck: - `HasAssignments` is a collection, some sort of `[1]` accessor is missing.
- `IfcRelAssignsToProduct.RelatingProduct` is a select, a cast is missing.

pjanck (changes requested):

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
SergejMuhic: Has been removed.

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcIntegerVoxelData/DocEntity.xml
SergejMuhic: Not sure I want to do that. Keeping consistency here.

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcVoxelData/DocEntity.xml
SergejMuhic: Rule has been reworked to have domain and @larswik requirements aligned.

PR #533, opened 2023-03-07
spiral dimensions in IfcCurveDim function

SergejMuhic:
Requirement documented here https://github.com/buildingSMART/IFC4.3.x-development/issues/571

1. [spiral dimensions in IfcCurveDim function](https://github.com/bSI-InfraRoom/IFC-Specification/commit/6e327b1b3ec3f5b6ce2d5f8668ade29cbf4140e3)


pjanck (approved):
lftm

peterrdf (approved):

iegorychev:
Agree with update.

However I'd introduce an optimizing suggestion for this and similar code in future.

TYPEOF(...) operation is quite hard actually on each step, so what if (not correct EXPRESS here, but idea):

VAR typepath := TYPEOF(Curve);

and then sequence of ifs:

if ('IFCGEOMETRYRESOURCE.IFCSPIRAL' IN typepath)

?

PS: This probably depends on expressions 'player' implementation, if SET of type names is cached or not, but with such optimization this kind of caching is explicit.

larswik (approved):

iegorychev (approved):

SergejMuhic:
> Agree with update.
> 
> However I'd introduce an optimizing suggestion for this and similar code in future.
> 
> TYPEOF(...) operation is quite hard actually on each step, so what if (not correct EXPRESS here, but idea):
> 
> VAR typepath := TYPEOF(Curve);
> 
> and then sequence of ifs:
> 
> if ('IFCGEOMETRYRESOURCE.IFCSPIRAL' IN typepath)
> 
> ?
> 
> PS: This probably depends on expressions 'player' implementation, if SET of type names is cached or not, but with such optimization this kind of caching is explicit.

In this case I would go for a switch instead of individual ifs.

iegorychev:
> > Agree with update.
> > However I'd introduce an optimizing suggestion for this and similar code in future.
> > TYPEOF(...) operation is quite hard actually on each step, so what if (not correct EXPRESS here, but idea):
> > VAR typepath := TYPEOF(Curve);
> > and then sequence of ifs:
> > if ('IFCGEOMETRYRESOURCE.IFCSPIRAL' IN typepath)
> > ?
> > PS: This probably depends on expressions 'player' implementation, if SET of type names is cached or not, but with such optimization this kind of caching is explicit.
> 
> In this case I would go for a switch instead of individual ifs.

Seems not convenient as operator IN here. It should find a line within SET on each case. Would be possible in case 'SCHEMANAME.ENTITY_NAME' = nameToFind

SergejMuhic:
True, my bad.

SergejMuhic:
I had an idea. What do you think about this:
```
CASE TRUE OF
    VALUE_IN(typepath,'IFCGEOMETRYRESOURCE.IFCSPIRAL') : ;
```

iegorychev:
Hm... this is allowed, and the case with just IN is allowed too. So my bad here, EXPRESS allows 'expressions' in case actions. =)

SergejMuhic:
yes of course, `IN` operator would work in this format too.

PR #547, opened 2023-03-07
Changed the wording of General Usage for Project Template Definitions

### LINKED ISSUE #546: Wording of Project Template Definitions usage on IfcProject
https://github.com/bSI-InfraRoom/IFC-Specification/blob/3da5c0a79c84199a649dd10086b6f2834a11d90b/IFC4x3/ModelViews/General%20Usage/DocModelView.xml#L10398

I believe we should revise this.

pjanck:
I am unsure about this change. But I believe it is more clear now, what is the requirement.

I would love to hear your opinions.

Fixes #546 .


SergejMuhic:
I dunno. Maybe a compromise, just remove the "for"?

larswik (approved):

PR #563, opened 2023-04-13
Feature/wr alignment representation

SergejMuhic:
1. [wr with function for gradient curve](https://github.com/bSI-InfraRoom/IFC-Specification/commit/8d3e0082f34f6fb40639f3e6dd5a532d5f1420e5)

>NOTE 1: no segmented reference curve rule yet

>NOTE 2: do we want all valid representations (polyline, indexedPolycurve, offsetCurveByDistances etc.) in this rule?


IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Functions/IfcAlignmentRepresentation/Expression.txt
```diff
@@ -0,0 +1,40 @@
+(Alignment : IfcAlignment) : IfcBoolean;
+	
+	LOCAL
+		AlignmentRepresentations : SET OF IfcShapeRepresentation := SELF.Representation.Representations;
+	END_LOCAL;
```
pjanck: I wonder how this will be initialized if `SELF.Representation` is `NIL`. (it is checked only after these lines)

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Functions/IfcAlignmentRepresentation/Expression.txt
```diff
@@ -0,0 +1,40 @@
...
+	IF SIZEOF(AlignmentRepresentations) = 1
+	AND
+	(SIZEOF(QUERY(twoDee <* AlignmentRepresentations | twoDee.RepresentationType = 'Curve2D')) = 1
+	OR
+	SIZEOF(QUERY(threeDee <* AlignmentRepresentations | threeDee.RepresentationType = 'Curve3D')) = 1)
+		RETURN TRUE;
+	END_IF;
+	
+	/* 2 representations and 3D BaseCurve reference equals 2D representation */
+	IF SIZEOF(AlignmentRepresentations) = 2)
```
pjanck:
```suggestion
	IF (SIZEOF(AlignmentRepresentations) > 1)
```

- There may be more and not fail, right?
- A parenthesis is missing.

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Functions/IfcAlignmentRepresentation/Expression.txt
```diff
@@ -0,0 +1,40 @@
...
+	
+		TwoDRep : SET OF IfcRepresentationItem := QUERY(twoDee <* AlignmentRepresentations | twoDee.RepresentationType = 'Curve2D');
+		ThreeDRep : SET OF IfcRepresentationItem := QUERY(threeDee <* AlignmentRepresentations | threeDee.RepresentationType = 'Curve3D');
+		IF SIZEOF(ThreeDRep) = 1 AND ('IFCGEOMETRYRESOURCE.IFCGRADIENTCURVE' IN TYPEOF(ThreeDRep[1].Items[1]))
+			IF SIZEOF(TwoDRep) = 1 AND ('IFCGEOMETRYRESOURCE.IFCCOMPOSITECURVE' IN TYPEOF(TwoDRep[1].Items[1]))
```
pjanck: Should we query instead of `[1]`, in case there are multiple `Items`?

pjanck (changes requested):
Initial comments.

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Functions/IfcAlignmentRepresentation/Expression.txt
SergejMuhic: I have to look up how local variables are initialized. I did not think before initializing. Typically they are first in the function. 😄

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Functions/IfcAlignmentRepresentation/Expression.txt
SergejMuhic: I would not allow more, hence return false at the end. If there are more, what is used for positioning?

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Functions/IfcAlignmentRepresentation/Expression.txt
SergejMuhic: Yes and no. A query also returns a SET of 1.

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Functions/IfcAlignmentRepresentation/DocFunction.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<DocFunction xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" Name="IfcAlignmentRepresentation" UniqueId="82d30aab-1426-4d44-8137-57335d7dfeaf">
+	<Localization>
+		<DocLocalization Locale="en">
+			<Documentation>Allignment representation.</Documentation>
```
pjanck:
```suggestion
			<Documentation>Alignment representation.</Documentation>
```

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Entities/IfcAlignment/DocEntity.xml
```diff
@@ -8,5 +8,12 @@
 			<Documentation></Documentation>
 		</DocAttribute>
 	</Attributes>
+	<WhereRules>
+		<DocWhereRule Name="AlignmentRepresentation" UniqueId="487a34c4-a3b3-407c-8fbb-88e476a16737">
+			<Documentation>Only _IfcShapeRepresentation_.RepresentationIdentifier=&apos;Axis&apos; and _IfcShapeRepresentation_.RepresentationType=&apos;Curve2D&apos; or &apos;Curve3D&apos; are allowed. If both &apos;Curve2D&apos; and &apos;Curve3D&apos; are defined, then the horizontal alignment 2D curve shall be the basis of the Curve3D.
```
pjanck:
```suggestion
			<Documentation>Only _IfcShapeRepresentation_.RepresentationIdentifier=&apos;Axis&apos; and _IfcShapeRepresentation_.RepresentationType=&apos;Curve2D&apos; or &apos;Curve3D&apos; are allowed. If both &apos;Curve2D&apos; and &apos;Curve3D&apos; are defined, then the horizontal alignment 2D curve shall be the basis of the 3D curve.
```

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Functions/IfcAlignmentRepresentation/Expression.txt
```diff
@@ -0,0 +1,42 @@
...
+	END_IF;
+	
+	(* 2 representations and 3D BaseCurve reference equals 2D representation *)
+	IF SIZEOF(AlignmentRepresentations) = 2 THEN
+	
+		TwoDRep := QUERY(twoDee <* AlignmentRepresentations | twoDee.RepresentationType = 'Curve2D');
+		ThreeDRep := QUERY(threeDee <* AlignmentRepresentations | threeDee.RepresentationType = 'Curve3D');
+		IF SIZEOF(ThreeDRep) = 1 AND ('IFCGEOMETRYRESOURCE.IFCGRADIENTCURVE' IN TYPEOF(ThreeDRep[1].Items[1])) THEN
+			IF SIZEOF(TwoDRep) = 1 AND ('IFCGEOMETRYRESOURCE.IFCCOMPOSITECURVE' IN TYPEOF(TwoDRep[1].Items[1])) THEN
+				IF ThreeDRep[1].Items[1]\IfcGradientCurve.BaseCurve :=: TwoDRep[1].Items[1] THEN
```
pjanck: What if there are multiple `.Items`?

pjanck (changes requested):

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Functions/IfcAlignmentRepresentation/Expression.txt
SergejMuhic: How about a check before:
```
IF SIZEOF(ThreeDRep[1].Items) > 1 THEN RETURN(FALSE);
```
?

pjanck: Another option would be to query and save as local variable:

> curveThreeD := QUERY(itemThreeD <* ThreeDRep[1].Items | 'IFCGEOMETRYRESOURCE.IFCGRADIENTCURVE' in TYPEOF(itemThreeD) );

Similarly for 2D representation (`curveTwoD`).

And then test:

> RETURN SIZEOF(curveThreeD) = 1 AND SIZEOF(curveTwoD) = 1 AND curveThreeD\IfcGradientCurve.BaseCurve :=: curveTwoD;

I'm unsure about the syntax, but hopefully I manage to get the idea across.

pjanck (commented):
A counter-proposal.

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Functions/IfcAlignmentRepresentation/Expression.txt
SergejMuhic: Mhm, I like that idea. Not sure if I have the time to do it before Thursday though. 😟

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Functions/IfcAlignmentRepresentation/Expression.txt
```diff
@@ -15,28 +17,39 @@
...
+		END_IF;
+		IF SIZEOF(QUERY(threeDee <* AlignmentRepresentations | threeDee.RepresentationType = 'Curve3D')) = 1
+				'IFCGEOMETRYRESOURCE.IFCGRADIENTCURVE' IN TYPEOF(AlignmentRepresentations[1].Items[1]):
+					RETURN(TRUE);
+				'IFCGEOMETRYRESOURCE.IFCOFFSETCURVEBYDISTANCE' IN TYPEOF(AlignmentRepresentations[1].Items[1]):
+					RETURN(TRUE);
+				'IFCGEOMETRYRESOURCE.IFCSEGMENTEDREFERENCECURVE' IN TYPEOF(AlignmentRepresentations[1].Items[1]):
+					RETURN(TRUE);
+				'IFCGEOMETRYRESOURCE.IFCPOLYLINE' IN TYPEOF(AlignmentRepresentations[1].Items[1]):
+					RETURN(TRUE);
```
pjanck: `CASE TRUE OF` missing.

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Functions/IfcAlignmentRepresentation/Expression.txt
```diff
@@ -15,28 +17,39 @@
...
+		END_IF;
```
pjanck: `END_CASE` missing for both.

I wonder about `DEFAULT` - should it return `UNKNOWN`?

pjanck: This assumes that `SIZEOF(AlignmentRepresentations[1].Items) = 1`. I wonder if such assumption is correct.

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Functions/IfcAlignmentRepresentation/Expression.txt
```diff
@@ -15,28 +17,39 @@
...
-		IF SIZEOF(ThreeDRep) = 1 AND ('IFCGEOMETRYRESOURCE.IFCGRADIENTCURVE' IN TYPEOF(ThreeDRep[1].Items[1])) THEN
-			IF SIZEOF(TwoDRep) = 1 AND ('IFCGEOMETRYRESOURCE.IFCCOMPOSITECURVE' IN TYPEOF(TwoDRep[1].Items[1])) THEN
-				IF ThreeDRep[1].Items[1]\IfcGradientCurve.BaseCurve :=: TwoDRep[1].Items[1] THEN
-					RETURN(TRUE);
-				END_IF;
-			END_IF;
-		END_IF;
+		CurveThreeD := QUERY(item <* ThreeDRep[1].Items | 'IFCGEOMETRYRESOURCE.IFCGRADIENTCURVE' in TYPEOF(item));
+		
+		RETURN(SIZEOF(CurveThreeD) = 1 AND SIZEOF(CurveTwoD) = 1 AND CurveThreeD\IfcGradientCurve.BaseCurve :=: CurveTwoD);
```
pjanck, line 56: Is it necessary to nominate `\IfcGradientCurve` to access `BaseCurve` attribute? The `CurveThreeD` variable is already of type `IfcGradientCurve`.

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Functions/IfcAlignmentRepresentation/Expression.txt
```diff
@@ -15,28 +17,39 @@
...
 	
 	END_IF;
+	(* Missing an IFCSEGMENTEDREFERENCECURVE various placement option implementeation. *)
```
pjanck:
```suggestion
	(* Missing an IFCSEGMENTEDREFERENCECURVE various placement option implementation. *)
```

pjanck (changes requested):

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Functions/IfcAlignmentRepresentation/Expression.txt
SergejMuhic: I am just following suit from here: https://github.com/AECgeeks/infra-repo-issue-test-2/issues/16
since I did not get a definitive answer on my examples listed,

I did not understand the logic behind the change. It seems to me that p11 would see more importance on the IfcAxis2Placement example:
"NOTE This usage is required when a complex entity instance type has multiple attributes with the
same name or when a select data type contains multiple entities with attributes having the same name."

or even more important here:
```
LocationIs3D : SELF\IfcPlacement.Location.Dim = 3;
```
as Dim is not defined on IfcPoint

or here:
```
(Arg1.Dim <> Arg2.Dim)
```
where Arg1 and Arg2 are IfcVectorOrDirection where both entities have the same Dim attribute name.

I guess we are trusting someone more knowledgeable than us ->https://github.com/AECgeeks/infra-repo-issue-test-2/issues/16#issuecomment-1257575294 "but I trust the user of the tool to be highly knowledgeable."

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Functions/IfcAlignmentRepresentation/Expression.txt
```diff
@@ -0,0 +1,58 @@
...
+			CASE TRUE OF
+				'IFCGEOMETRYRESOURCE.IFCCOMPOSITECURVE' IN TYPEOF(AlignmentRepresentations[1].Items[1]):
+					RETURN(TRUE);
+				'IFCGEOMETRYRESOURCE.IFCOFFSETCURVEBYDISTANCE' IN TYPEOF(AlignmentRepresentations[1].Items[1]):
+					RETURN(TRUE);
+				'IFCGEOMETRYRESOURCE.IFCPOLYLINE' IN TYPEOF(AlignmentRepresentations[1].Items[1]):
+					RETURN(TRUE);
+			END_CASE;
+		END_IF;
+		IF SIZEOF(QUERY(threeDee <* AlignmentRepresentations | threeDee.RepresentationType = 'Curve3D')) = 1
```
pjanck:
```suggestion
		IF SIZEOF(QUERY(threeDee <* AlignmentRepresentations | threeDee.RepresentationType = 'Curve3D')) = 1 THEN
```

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Functions/IfcAlignmentRepresentation/Expression.txt
```diff
@@ -0,0 +1,58 @@
...
+	END_IF;
+	
+	(* Alignment.Representation exists so assign *)
+	AlignmentRepresentations := Alignment.Representation.Representations;
+
+	(* one representation is either Curve2D or 3D *)
+	IF SIZEOF(AlignmentRepresentations) = 1 THEN
+		IF SIZEOF(QUERY(twoDee <* AlignmentRepresentations | twoDee.RepresentationType = 'Curve2D')) = 1 THEN
+			CASE TRUE OF
+				'IFCGEOMETRYRESOURCE.IFCCOMPOSITECURVE' IN TYPEOF(AlignmentRepresentations[1].Items[1]):
```
pjanck, line 26:
```suggestion
				SIZEOF( QUERY(item <* AlignmentRepresentations[1].Items | 'IFCGEOMETRYRESOURCE.IFCCOMPOSITECURVE' in TYPEOF(item)) = 1:
```

Similar for other cases. (I'm unsure in the syntax, please double check.)

pjanck (changes requested):

IFC4x3/Sections/Core data schemas/Schemas/IfcProductExtension/Functions/IfcAlignmentRepresentation/Expression.txt
SergejMuhic: I would prefer something like:
```
SIZEOF(QUERY(... IN TYPEOF(...))) = SIZEOF(AlignmentRepresentations.Items)
```
if we want to go down that route.

SergejMuhic: After discussion with @michel-rives another proposal that makes sense from the domain and data exchange point of view. Before the case statement a check that
```
SIZEOF(AlignmentRepresentations[1].Items) = 1
```
else fail.

pjanck (approved):
Seems to accomplish its job!

PR #580, opened 2023-04-20
BookA-Pset_GeoObsBoreholeCommon

larswik:
# Pset: Pset_GeoObsBoreholeCommon

## Description: 

Common properties for borehole logs

|BoreholeName|IfcLabel|Identifier of a borehole like No. and name (as same as Pset_SpatialGeoObsBorehole 'BoreholeName' - Not part of the Pset/Qto|
|DepthRange|IfcNonNegativeLengthMeasure|Upper and bottom depth for certain geological information|


IFC4x3/Properties/d/DepthRange_0NQ19_ZHX2QfMh8143pjcc/DocProperty.xml
SergejMuhic, line 1: Is this not possible to derive from geometry?

j-weil: The intention was:
The depth range is what you get out of a borehole database and we want to see it as a property as well, very important. This Pset "GeoObsBoreholeCommon" includes nothing else because there are more Psets for different kinds of logs:
image

j-weil: Kazunori to check datatype @kazunorit

mbeaufils:
Regarding the DepthRange, the intention is to have an interval. Ex : [10m ; 35m]. Could you @larswik , @SergejMuhic confirm that IfcNonNegativeLengthMeasure is ok for that?

larswik:
> Regarding the DepthRange, the intention is to have an interval. Ex : [10m ; 35m]. Could you @larswik , @SergejMuhic confirm that IfcNonNegativeLengthMeasure is ok for that?

So, if I understand correctly, DepthRange should be a bounded value with primary datatype IfcNonNegativeLengthMeasure. this is ok. The property type needs to change from single value to bounded value. Datatype stays the same.

larswik:
BoreholeName is handled by IfcGeoScienceObservation.Name thus not included in this Pset.

IFC4x3/Properties/d/DepthRange_0NQ19_ZHX2QfMh8143pjcc/DocProperty.xml
kazunorit: @j-weil @larswik  DepthRange is basically a range of distance from the head, so I think IfcNonNegativelengthMeasure is fine. Please make sure its property type is a bounded value.

larswik: > @j-weil @larswik DepthRange is basically a range of distance from the head, so I think IfcNonNegativelengthMeasure is fine. Please make sure its property type is a bounded value.

Done. See latest commit.

SergejMuhic (approved):

PR #585, opened 2023-04-20
BookA-Pset_GeoObsBoreholeWaterLevel

larswik:
# Pset: Pset_GeoObsBoreholeWaterLevel

## Description: 

Properties for describing water levels

|WaterLevel|IfcLengthMeasure|measured distance to water level in borehole from ground surface, in meters|
|IsPiezometric|IfcBoolean|Defines whether hight measurements are made vs the piezometric surface (true) or the surface of the groundwater (false).|


IFC4x3/Properties/w/WaterLevelType_1c5OF998b1_8SQzyRB0_vU/Documentation.md
```diff
@@ -0,0 +1 @@
+Type of measured water level (free groundwater / rise to this measured level).
```
SergejMuhic: Having trouble understanding this description. What would be an example of a `WaterLevelType`?

Cannot make suggestion if I do not understand.

larswik: > Having trouble understanding this description. What would be an example of a `WaterLevelType`?
> 
> Cannot make suggestion if I do not understand.

Ping @j-weil.

j-weil: Water level types would be
"free" or "piezometric / confined" when the groundwater body is limited on top by a seal and the water will rise in a pipe that sticks into it. Helpful?

larswik: proposal: IsPiezometric : IfcBoolean - Defines whether hight measurements are made vs the piezometric surface (true) or the surface of the groundwater (false)

Found this definition of piezometric surface: "the imaginary surface to which groundwater rises. under hydrostatic pressure in wells or spring."

@j-weil: Please confirm!

larswik: Fixed in [1b1bc72](https://github.com/bSI-InfraRoom/IFC-Specification/pull/585/commits/1b1bc72aa23d64541fd411ac8250a25a69c46a2d)

larswik:
Also saw that WaterLevelMbgS now should be just WaterLevel. Fixed in [1b1bc72](https://github.com/bSI-InfraRoom/IFC-Specification/pull/585/commits/1b1bc72aa23d64541fd411ac8250a25a69c46a2d)

IFC4x3/Properties/i/IsPiezometric_1c5OF998b1_8SQzyRB0_vU/Documentation.md
```diff
@@ -0,0 +1 @@
+Defines whether hight measurements are made vs the piezometric surface (true) or the surface of the groundwater (false).
```
SergejMuhic: Typo carried over from Excel. 😃 
```suggestion
Defines whether height measurements are made vs the piezometric surface (true) or the surface of the groundwater (false).
```

SergejMuhic (approved):

PR #607, opened 2023-04-25
BookA-Pset_InSituTestCommon

larswik:
# Pset: Pset_InSituTestCommon

## Description: 

Properties in common for in-situ tests in soil or rock.

|DepthRange|IfcNonNegativeLengthMeasure|The depth range where the test was performed.|
|TestMethod|IfcLabel|The test method used.|
|TestType|IfcLabel|The test type.|


patmclarin:
please update the Pset description to: Properties in common for in-situ tests in soil or rock

PR #611, opened 2023-04-25
BookA-Pset_LabTestCommon

larswik:
# Pset: Pset_LabTestCommon

## Description: 

Properties describing the laboratory test used to obtain the associated results.

|TestMethod|IfcLabel|Name of the test method that was used to optain the exchanged results.|
|TestStandard|IfcLabel|Reference to standard that describes the test merthodology.|
|TestType|IfcLabel|Discipline responsible for performing the tests and using ist results.|


patmclarin:
can we update the Pset description to: Properties describing the laboratory test used to obtain the associated results

PR #614, opened 2023-04-28
BookB-Pset_GeoScienceFeatureCommon

larswik:
# Pset: Pset_GeoScienceFeatureCommon

## Description: 

Common properties for geoscience features

|Comment|IfcText|Additional noteworthy information related e.g. to one specific entity|

### DIFF
--- a/IFC4x3/Properties/c/Comment_3bMkv0QVr4JxAPUSABd8Y9/DocProperty.xml
+++ b/IFC4x3/Properties/c/Comment_3bMkv0QVr4JxAPUSABd8Y9/DocProperty.xml
@@ -1,0 +1,2 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<DocProperty xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" id="Comment_3bMkv0QVr4JxAPUSABd8Y9" Name="Comment" UniqueId="e55aee40-69fd-444f-b299-79c28b9c8889" PrimaryDataType="IfcText" />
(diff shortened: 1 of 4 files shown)


SylvieBretelle:
more properties into this GeoScienceFeatureCommon : description, comment, FeatureName, linktodata, ProjectName, Chainage

larswik:
> more properties into this GeoScienceFeatureCommon : description, comment, FeatureName, linktodata, ProjectName, Chainage

See comments in Excel sheet:
Description => IfcGeoScienceFeature.Descriptioin
Comment => Is already there (see above)!
FeatureName => IfcGeoScienceFeature.Name
linkToData => IfcRelAssociatesDataset
ProjectName => IfcProject.Name
Chainage => IfcRelPositions => IfcReferent/Pset_Stationing

=> Use mechanisms already in the schema instead of creating parallel mechanisms in Psets!

SergejMuhic:
Quite confused with this pset. None of the common properties e. g. `Reference` in it but one peculiar one `Comment` which seems a lot like `IfcRoot`.`Description`:

"Optional description, provided for exchanging informative comments."

larswik:
> Quite confused with this pset. None of the common properties e. g. `Reference` in it but one peculiar one `Comment` which seems a lot like `IfcRoot`.`Description`:
> 
> "Optional description, provided for exchanging informative comments."

Well, they had also a Description property specified, which I mapped to IfcRoot.Description so that one was already taken. Description from the domain team: "Any noteworthy items which could not be entered among other properties".
Comment has the description above (in PR description) from the domain team.

Regarding "Reference", we need to ask the domain team. To me it feels as if the standard properties "Reference" and "Status" seems to have more to do with man made objects.

Ping @j-weil



</body>

</html>

larswik:
If Description and Comment is one and the same property, we could skip this Pset altogether since everything else is already in the IFC schema.

j-weil:
I agree, if only Decription and Comment remain and this is already handled mor egenerally we do not need this Pset.

PR #615, opened 2023-04-28
BookB-Pset_GeoScienceFeatureTypeContact

larswik:
# Pset: Pset_GeoScienceFeatureTypeContact

## Description: 

Defines properties for contacts where a contact is a general concept representing any kind of surface separating two geologic units, including primary boundaries such as depositional contacts, all kinds of unconformities, intrusive contacts, and gradational contacts, as well as faults that separate geologic units. 

|GeoContactType|IfcIdentifier|Reports a URI referring to a controlled concept from a vocabulary defining the contact types.|
|DipDirection|IfcPlaneAngleMeasure|The geographic  direction of a line created by the intersection of a plane and the horizontal plane. If non specific convention is used, this angular value is in the range 0 to 18 degrees. In this case, there is an ambiguity on the dip orientation. For example, a plane with an orientation of 90 degrees from the nord could either have a dip direction to the North or to the South. Thus, to avoid this ambiguity, a strike value is generally completed with an indication of the dip orientation (Quadrant).|


SylvieBretelle:
proposed description:
A contact is a general concept representing any kind of surface separating two geologic units, including primary boundaries such as depositional contacts, all kinds of unconformities, intrusive contacts, and gradational contacts, as well as faults that separate geologic units. The contact is described with the following specific properties.

larswik:
> proposed description: A contact is a general concept representing any kind of surface separating two geologic units, including primary boundaries such as depositional contacts, all kinds of unconformities, intrusive contacts, and gradational contacts, as well as faults that separate geologic units. The contact is described with the following specific properties.

Since we are describing the property set (and not the contexct itself), I propose a slight change to:
Defines properties for contacts where a contact is a general concept representing any kind of surface separating two geologic units, including primary boundaries such as depositional contacts, all kinds of unconformities, intrusive contacts, and gradational contacts, as well as faults that separate geologic units.

SylvieBretelle:
perfect to me !


Sylvie Bretelle


________________________________
De : larswik ***@***.***>
Envoyé : jeudi 4 mai 2023 14:16
À : bSI-InfraRoom/IFC-Specification ***@***.***>
Cc : SylvieBretelle ***@***.***>; Comment ***@***.***>
Objet : Re: [bSI-InfraRoom/IFC-Specification] BookB-Pset_GeoScienceFeatureTypeContact (PR #615)


proposed description: A contact is a general concept representing any kind of surface separating two geologic units, including primary boundaries such as depositional contacts, all kinds of unconformities, intrusive contacts, and gradational contacts, as well as faults that separate geologic units. The contact is described with the following specific properties.

Since we are describing the property set (and not the contexct itself), I propose a slight change to:
Defines properties for contacts where a contact is a general concept representing any kind of surface separating two geologic units, including primary boundaries such as depositional contacts, all kinds of unconformities, intrusive contacts, and gradational contacts, as well as faults that separate geologic units.

PR #632, opened 2023-05-08
BookB-Pset_GeoSciencePropertiesCommon

larswik:
# Pset: Pset_GeoScienceProperties

## Description: 

Properties common for geological and geotechnical units.

|PwaveVelocity|IfcLinearVelocityValue|Value of S wave (Shear wave (S)) measured through the unit.|
|SwaveVelocity|IfcLinearVelocityValue|Value of P wave (pressure or primary elastic body wave (P)) measured through the unit.|
|Resistivity| IfcElectricalResistivityMeasure | Electrical resistivity of a rock or soil (Ohm-m). |
|GroundwaterTemperature|IfcThermodynamicTemperatureMeasure|Temperature measured or assumed in the groundwater.|
|WetBulkDensity|IfcMassDensityMeasure|Ratio of the total mass to the unit total volume (material at its natural moisture content).|
|DryBulkDensity|IfcMassDensityMeasure|Ratio of the mass of solid to the unit total volume (dry material).|


kazunorit:
Checked datatypes for Resistivity at:
https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/ifcmeasureresource/content.html

Can't find an appropriate datatype there. Needs to add a new one.
Also, better to re-name it to "Electrical Resistivity" to make it clear and to match an existing "Electric Resistance" (I think the correct term is "ELECTRICAL resistance", not electric).

larswik:
> Checked datatypes for Resistivity at: https://ifc43-docs.standards.buildingsmart.org/IFC/RELEASE/IFC4x3/HTML/ifcmeasureresource/content.html
> 
> Can't find an appropriate datatype there. Needs to add a new one. Also, better to re-name it to "Electrical Resistivity" to make it clear and to match an existing "Electric Resistance" (I think the correct term is "ELECTRICAL resistance", not electric).

Just for clarity: are we speaking ElectricalResistance (Ohm - measure type exists in IFC as IfcElectricResistanceMeasure ) or ElectricalResistivity (Ohm-meter - no predefined measure type exists currently in IFC)?
Resistivity was used in my proposal because there is already a defined property in the spec (used by Pset_ProtectiveDeviceTypeSparkGap and Pset_SolidStratumCapacity).

larswik:
ping @kazunorit : Can you check my last comment?

kazunorit:
@larswik 
I see that Pset_ProtectiveDeviceTypeSparkGap and Pset_SolidStratumCapacity have the property "Resistivity" of which the data type is IfcElectricResistanceMeasure. "Resistance" and "resistivity" are different physical properties by definition, and IfcElectricalResistanceMeasure comes with the unit "Ohm" whilst "resistivity" is in Ohm-m. The question is how precisely properties need to be defined/used in IFC but, as an electrical engineer, I feel clearly distinguishing these two properties is the right thing to do, which requires adding another property "resistivity" apart from "resistance".

SergejMuhic:
@kazunorit properties in IFC are as precise as you want them since these are metadata on objects.

Tolerances, data quality etc. are part of MVDs. We are not looking into that yet. The most important job atm is to have a framework that specifies the datastructures and fields with their appropriate type.

larswik:
Proposal:
- Add type IfcElectricalResistivityMeasure (unit: ohm-m)
- Change type of Resistivity property to the above.
- Add resistance property (type: IfcElectricResistanceMeasure, unit: ohm)
- And check the Pset_ProtectiveDeviceTypeSparkGap (resistivity property)!!

larswik:
> * Pset_ProtectiveDeviceTypeSparkGap

When googling, I only find writing about resistance for spark gaps. Furthermore, the definition of resistivity "Electrical resistivity of a rock or soil (ohm-m)" seems misplaced for a spark gap (to say the least). SPARKGAP was added as a PDT in IFC 4x3 and my guess that it should be Resistance rather than Resistivity.

SergejMuhic:
> > * Pset_ProtectiveDeviceTypeSparkGap
> 
> When googling, I only find writing about resistance for spark gaps. Furthermore, the definition of resistivity "Electrical resistivity of a rock or soil (ohm-m)" seems misplaced for a spark gap (to say the least). SPARKGAP was added as a PDT in IFC 4x3 and my guess that it should be Resistance rather than Resistivity.

In this case, let us propose to replace the `Resistivity` property with the added `Resistance` property.

larswik:
> 4020b6e

Fixed in [4020b6e](https://github.com/bSI-InfraRoom/IFC-Specification/pull/632/commits/4020b6ec1cd46ce61f18eb12a0ea04418fec45b4)

PR #564, opened 2023-04-14
Pset_EarthworksFillCommon

larswik:
# Pset: Pset_EarthworksFillCommon

## Description: 

Common properties for earthworks fill elements

| Property              | Datatype                | Description                                                  |
| --------------------- | ----------------------- | ------------------------------------------------------------ |
| SourceOfFill          | PEnum_SourceOfFill      | Source of fill                                               |
| SourceOfFillOther     | IfcLabel                | User defined Source of fill when SourceOfFill = OTHER        |
| CompactionMethod      | PEnum_CompactionMethod  | Method of compaction                                         |
| CompactionMethodOther | IfcLabel                | User defined method of compaction when CompactionMethod = OTHER |
| CompactionFactor      | IfcPositiveRatioMeasure | Reduction in volume on compaction                            |

## Enumeration: PEnum_SourceOfFill

| Constant  | Description |
| --------- | ----------- |
| IMPORTED  |The fill material is imported.|
| EXCAVATED |The fill material is excavated on site.|
| OTHER     |Some other source is used for the fill. Use property SourceOfFillOther to describe the source. |

## Enumeration: PEnum_CompactionMethod

| Constant | Description |
| -------- | ----------- |
| ROLLER   |Compaction using some type of roller equipment.|
| TAMPING  |Compaction using some kind of tamping equipment.|
| NONE     |No compaction.|
| OTHER    |Some other means of compaction. Use property CompactionMethodOther to describe the means.|

PR #675, opened 2023-11-26
error in DocEntity IfcGeoscienceElementType

SergejMuhic:
Was not opening. Did not check where the bug is from, just corrected.


larswik (commented):
I fixed this in #564.

SergejMuhic:
> I fixed this in #564.

So, I have to merge #564 first then?

I would prefer to keep the PRs clear i. e. merge this one and after this the psets.

PR #734, opened 2024-02-12
Material usage for IfcImprovedGround

larswik:


pjanck (approved):
Quite some noise in the diff with content that moved around in the serialized view. But still manageable.

The added content for `IfcImprovedGround`: LFTM.

SergejMuhic (approved):

PR #803, opened 2024-05-10
Pset 2nd reconciliation

larswik:
Doing all the remaining stuff, e.g.:
- Deprecating and unifying the Reference property
- Correcting other errors that occurred during 1st reconciliation round


larswik:
Done everything except deleting Psets that seem to have been deleted in 4.3.

PR #808, opened 2024-05-27
Reconcile/ifc point dim

### LINKED ISSUE #767: [BucketA]Reconcile IfcPoint.Dim
The proper modelling technique for this should not be a function but redeclared attributes.

SergejMuhic:
1. [moving DIm from IfcPoint subtypes to IfcPoint](https://github.com/bSI-InfraRoom/IFC-Specification/commit/8d4ada4c9e372511eb53f875f314be3c8ca47f50)

2. [adding IfcPointDim function](https://github.com/bSI-InfraRoom/IFC-Specification/commit/4c97ec01721d44c4939b44e0096b99338a6910e3)


larswik (approved):

iegorychev:
In this case DERIVED attribute which is called Dim for derived classes can be compared with virtual methods of C++ when every class knows how to determine Dim for itself, while the way of function calling looks like non-virtual method of base class dynamically casts itself to derived classes to determine Dim. Looks heavy at least because of function call and three TYPEOF(Point) calls in worst case. Why not the old way?

SergejMuhic:
> In this case DERIVED attribute which is called Dim for derived classes can be compared with virtual methods of C++ when every class knows how to determine Dim for itself, while the way of function calling looks like non-virtual method of base class dynamically casts itself to derived classes to determine Dim. Looks heavy at least because of function call and three TYPEOF(Point) calls in worst case. Why not the old way?

Yes, I completely agree. I never understood this change. I argued the exact same thing a bunch of times but cannot find the issues anymore. Maybe @aothms or @TLiebich remember?

It is an unnecessary change and should be reverted, especially, because it is a change for the worse.

PR #850, opened 2024-06-25
optional IfcCurveBoundedPlane.InnerBoundaries and cardinality 1 to many

### LINKED ISSUE #660: Non-optional SET [0:?]
Hi All,

ENTITY IfcCurveBoundedPlane
 SUBTYPE OF (IfcBoundedSurface);
	BasisSurface : IfcPlane;
	OuterBoundary : IfcCurve;
	InnerBoundaries : SET [0:?] OF IfcCurve;
END_ENTITY;

Validation "Validate required attributes assigned" should return TRUE, if all non-optional attributes have values, or if object has non-optional attributes [only]; FALSE if any non-optional attribute has no value (10303-22).

I've just got a file with InnerBoundaries=$ which doesn't pass validation as it should be () at least in correct case. =)

Shouldn't such cases for explicit attributes be transformed into InnerBoundaries : OPTIONAL SET [1:?] OF IfcCurve; ?

SergejMuhic:
closes #660 

implementing proposal from @iegorychev

### DIFF
--- a/IFC4x3/Sections/Resource definition data schemas/Schemas/IfcGeometryResource/Entities/IfcCurveBoundedPlane/DocEntity.xml
+++ b/IFC4x3/Sections/Resource definition data schemas/Schemas/IfcGeometryResource/Entities/IfcCurveBoundedPlane/DocEntity.xml
@@ -11,7 +11,7 @@
 		<DocAttribute Name="OuterBoundary" UniqueId="d59dedb7-409b-40f2-b7f6-780fbf45198e" DefinedType="IfcCurve" XsdFormat="attribute">
 			<Documentation>The outer boundary of the surface.</Documentation>
 		</DocAttribute>
-		<DocAttribute Name="InnerBoundaries" UniqueId="33e2f437-e02b-4345-af76-39640de58ff3" DefinedType="IfcCurve" AggregationType="3" AggregationLower="0" AggregationUpper="?">
+		<DocAttribute Name="InnerBoundaries" UniqueId="33e2f437-e02b-4345-af76-39640de58ff3" DefinedType="IfcCurve" AttributeFlags="1" AggregationType="3" AggregationLower="1" AggregationUpper="?">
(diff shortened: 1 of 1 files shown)


TLiebich (changes requested):
I would disagree to the proposed change, since it creates a non-upward compatible change in regard to IFC4.3. The SPF notation would change from [] "empty set" to $ "non asserted set" and at least under previous considerations this is not upward compatible.

(the original SET[0:?] also comes from P42).

Personally I would recommend to wait until the next major IFC release to make this kind of changes.

SergejMuhic:
> I would disagree to the proposed change, since it creates a non-upward compatible change in regard to IFC4.3. The SPF notation would change from [] "empty set" to $ "non asserted set" and at least under previous considerations this is not upward compatible.
> 
> (the original SET[0:?] also comes from P42).
> 
> Personally I would recommend to wait until the next major IFC release to make this kind of changes.

No worries, we will revert it for the submission. This was just for the test run. Thanks for the feedback.

I understand, the problem is that SPF serialization would suffer and old files that require an empty set would fail in deserialization since according to this spec an empty list is not allowed. A question though, what about OPTIONAL SET[0:?]. This way both $ and () are allowed.

A slight correction though, part 42 actually has SET[1:?]:
image since the definition is a bit different and there is no separate outer boundary attribute. One boundary has to be always in the set. I don't know, however, if this was changed or when it was changed but the current state in part 42 seems to be one mandatory SET[1:?].

IFC is also a bit inconsistent further, apparently two entities are referencing this one `curve_bounded_surface` definition from part 42:
1. [`IfcCurveBoundedSurface`](https://bsi-infraroom.github.io/IFC-Documentation-Tunnel/4_4_0_0/general/HTML/link/ifccurveboundedsurface.htm)
2. [`IfcCurveBoundedPlane`](https://bsi-infraroom.github.io/IFC-Documentation-Tunnel/4_4_0_0/general/HTML/link/ifccurveboundedplane.htm)

EDIT: Realized this one was not even merged. I was careful. Funny, I know.

PR #851, opened 2024-06-25
ifcsectionedsurface.areaprofiletypes

SergejMuhic:
fixes the WR and documentation to do what it is supposed to.


SergejMuhic:
merging for the purpose of the test with @aothms 

feel free to derive an issue to discuss it though. that is why  you all were tagged.

TLiebich (commented):
agree, it does now what it is supposed to do.

PS: whether or not this ProfileType enum makes sense in all occurrences within the IFC geometry is a different question, but this is rather an IFC5 (or successor) topic.

